PR #109, opened 2015-02-15
fuse_nfs: Disable multithreading

agraf:
If we blindly spawn new threads based on existing open NFS shares, all
threads will concurrently still run on the same file descriptors for NFS
traffic, colliding sooner or later.

This is what happens with the fuse-nfs example program. To steer us back
into safety, we can just tell fuse to not do multi-threading which this
patch does.

This patch fixes random connection aborts with fuse_nfs for me.

Signed-off-by: Alexander Graf <agraf@suse.de>


sahlberg:
merged, thanks

PR #118, opened 2015-06-19
For upstream

plieven:


sahlberg:
I don't think we should write to stderr from within libnfs since it is a library.
But as we don't have any logging subsystem yet ...

Can you remove the printf(stderr, calls for now,  then I can add a logging subsystem after next release.

plieven:
I added a patch to have a debug url parameter. libnfs produces no output to stderr if this is not set. This gives us the opportunity to have this supported in qemu 2.4.

PR #121, opened 2015-08-04
socket: handle count == 0 in rpc_read_from_socket

plieven:
An EOF is signalled through a POLLIN event and subsequen recvs return
always 0. Handle this condition and reconnect. Otherwise we might
deadlock here.

Signed-off-by: Peter Lieven <pl@kamp.de>


sahlberg:
Ah, yes.
We forgot to add this when we removed the ioctl(FIONREAD) call from the tcp path.

PR #122, opened 2015-08-06
socket: also reconnect on write errors

plieven:
Hi Ronnie,

I wonder if we should also reconnect on write errors?

Peter


plieven:
Ping

sahlberg:
Merged, Thanks

PR #128, opened 2015-09-10
Fuse symlinks

agraf:
With these commits I can successfully mount an nfs share on a remote system and chroot into it.


sahlberg:
Awesome!

Now we just need to convince these folks to add an entry for it:
http://sourceforge.net/p/fuse/wiki/FileSystems/

PR #131, opened 2016-01-08
Support Win32 compilation with Mingw

jbkempf:
This is a set of patches that should allow to cross-compile this library through Mingw on Linux

This is using autotools to compile for Win32, and not the win32build.bat

Some stuff are a bit kludgy to be honest, but the existing code was already a bit weird (nfs_stat declaration is fishy).

Some more work is required, I believe to clean everything for Win32, but this should be a good first step.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -91,6 +91,8 @@ fi
 AC_SUBST(WARN_CFLAGS)
 
 LIBSOCKET=
+SYS=
+
 case $host in
   *solaris*)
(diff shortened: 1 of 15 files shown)


jbkempf:
Ping :)

sahlberg:
Awesome!

Merged.

jbkempf:
Rocks :)

sahlberg:
Could you add a short sentence about MINGW to the README file too ?

PR #132, opened 2016-01-09
The correct NDK define for Android is __ANDROID__

jbkempf:
The ANDROID macro is not recommended and will not work in all cases.

This fixes the compilation for libnfs on Android.


sahlberg:
Thanks!

jbkempf:
Thanks for merging!

PR #133, opened 2016-01-14
Add a configure option to disable the utils compilation

fkuehne:
For some targets like iOS, utils are not useable, so it should be possible to disable their compilation

This is implemented in this pull request matching an existing configure switch to disable examples.


jbkempf:
Looks good to me :)

sahlberg:
Merged, Thanks!

PR #139, opened 2016-02-10
fix spelling errors.

doj:
fix some spelling errors in the function description of libnfs.h
Also state that the readahead is specified in bytes (I assume that after reading the code).


sahlberg:
Thanks!

PR #140, opened 2016-03-07
sync apis with timeout eg- nfs_write with nfs_write_t 

kkloginid:
Hi , was using this project and faced multiple time , thread hanging in nfs_write() api and few other because of network issues. because it was using -1 ( infinite ) poll time so threads are getting stuck forever. my write size is in chunks small sizes so i would expect them to finish in couple of second or just fail it. so added _t apis for all method , please see if it can be usefull for other people too and merge it in main code. existing methods are intact, so no issue with backward compatibility

### DIFF
--- a/lib/libnfs-sync.c
+++ b/lib/libnfs-sync.c
@@ -172,7 +172,7 @@ static void wait_for_nfs_reply_t(struct nfs_context *nfs, struct sync_cb_data *c
 		}
 		else if(retVal == 0)
 		{
-			nfs_set_error(nfs, "Timed out after [%ld] seconds",timeoutmiliseconds);
+			nfs_set_error(nfs, "Timed out after [%ld] miliseconds",timeoutmiliseconds);
 			cb_data->status = -EIO;
 			break;
 		}
--- a/lib/libnfs-sync.c
+++ b/lib/libnfs-sync.c
@@ -134,19 +134,49 @@ static void wait_for_reply(struct rpc_context *rpc, struct sync_cb_data *cb_data
 	}
 }
 
-static void wait_for_nfs_reply(struct nfs_context *nfs, struct sync_cb_data *cb_data)
+static void wait_for_nfs_reply(struct nfs_context *nfs, struct sync_cb_data *cb_data )
 {
 	struct pollfd pfd;
-
+	int retVal = -1;
 	while (!cb_data->is_finished) {
 
 		pfd.fd = nfs_get_fd(nfs);
 		pfd.events = nfs_which_events(nfs);
-		if (poll(&pfd, 1, -1) < 0) {
+		retVal = poll(&pfd, 1, -1);
+		if (retVal < 0) {
 			nfs_set_error(nfs, "Poll failed");
 			cb_data->status = -EIO;
 			break;
 		}
+		if (nfs_service(nfs, pfd.revents) < 0) {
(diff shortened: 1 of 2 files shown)


kkloginid:
fixed few things will add new rq closing this

sahlberg:
We definitely need support for setting timeouts in libnfs but I don't want
to add it this way by creating a new function for every single command.


I think it would be better to leave all the current functions as is and
instead control the timeout settings by adding new functions :
rpc_set_timeout(rpc, timeout)
nfs_set_timeout(nfs, timeout)

See libiscsi and do simething similar.

kkloginid:
Even my first thought was to have it like this and just modify wait_rpc , wait_nfs apis can solve this. but the doubt that i have is i might want just 2 seconds wait for nfs_create_dir  but if i am sending data in tune of 100 MBs i would like to have time out with bigger value , another problem is if i am using multiple threads to write files to nfs share ( same mount ) , it would affect each threads timeout , instead of passing in api as new arg.  But setting a quite reasonably big value would solve the infinite hang. eg: nfs_dir_create and nfs_write(100MB ) both having same time out 100 seconds. so its still better than not having any time out. please let me know if you want me to make these two api change and create a new pull req.

sahlberg:
On Wed, Mar 9, 2016 at 3:11 AM, Kuldeep Kumar <notifications@github.com>
wrote:

> Even my first thought was to have it like this and just modify wait_rpc ,
> wait_nfs apis can solve this. but the doubt that i have is i might want
> just 2 seconds wait for nfs_create_dir but if i am sending data in tune of
> 100 MBs i would like to have time out with bigger value , another problem
> is if i am using multiple threads to write files to nfs share ( same mount
> ) , it would affect each threads timeout , instead of passing in api as new
> arg. But setting a quite reasonably big value would solve the infinite
> hang. eg: nfs_dir_create and nfs_write(100MB ) both having same time out
> 100 seconds. so its still better than not having any time out. please let
> me know if you want me to make these two api change and create a new pull
> req.
>
libnfs is not threads safe so you can not safely share it between different
threads!

For multithreaded use you can either :
1, only ever call libnfs from one single dedicated thread.
2, create one dedicated context for each thread. As the contexts are very
cheap this is probably the best way.
3, wrap all calls into libnfs in a mutex.  This should be extremely cheap
no with the common support for futexes in linux.
As pretty much every single call into libnfs, except rpc/nfs_service, are
guaranteed to be non-blocking you are pretty much guaranteed
that the futex will almost never be contended so the cost would be near
zero.

When I initially wrote libnfs there were no futexes and I did not want to
use expensive mutexes to hurt performance.


The futex availability and that they are very cheap nowadays means that I
should probably add futex support to libnfs and make it thread safe without
having to worry about a potentially big impact on performance.




I think doing it the way libiscsi does it should address your cases.
Each pdu structure has a timeout field that is set when the pdu is
initialized.
It is initialized to the whatever value was set as the default on the
context in the most recent

It allows you to
1, set a default timeout for everything.   Just call nfs_set_timeout() once
immediately after creating the context.
This is likely what most people would use.

2, set a default timeout for almost everything and override it the few
times you wnat a different timeout.
I.e. call nfs_set_timeout() once on context initialization   then wrap any
calls you want to have a different timeout in :
  nfs_set_timeout(nfs, something-else)
  nfs_do_the_command()
  nfs_set_timeout(nfs, reset-to-default-again)

PR #141, opened 2016-03-07
sync apis with timeout value in milliseconds

kkloginid:
Was using this project and faced hang issues in few apis eg: nfs_write is hanging for infinite time as it is using poll time out value as -1 ( infinite ). 
added new apis to devloper like me to use with a timeout value to pass in each . depending upon the nature of api devloper should set proper timeout value. 
eg: for nfs_write_t(  ) it could be bit big value but for nfs_stat_t () it could be small 1 or 2 seconds.


kkloginid:
did you get time to see the latest change?

sahlberg:
The change was mainly good. Some style changes I fixed.

But the pull request was not good.
The series still contained the initial patches that added the _t functions and were later removed.
There were also several other patches that undid/redid things from previous patches.

For these things you should have a look at    git rebase -i 
That is an awesome command that allows you to merge multiple patches into one patch
(for example where you first create _t functions and then in a later patch remove them again).

For this time I did the rebase locally before I committed it.
But do have a look at git rebase -i   and git add   since these tools are great when re-working patches.


Thanks for the patch!
Next step for the timeouts should be to add support for timeouts for the async commands too.
1, Add a timeout field to the pdu structure too. Set this when the pdu is created from nfs_get_timeout().
2, See libiscsi for how we do timeouts for these functions.

I am not sure that returning -EIO for timeouts are the right thing to do. -ETIMEDOUT is probably a better errno value but the errno handling in libnfs is a bit dodgy already so ...

sahlberg:
I reworked the patch  and used git rebase to change it into a single patch
before merging.

Please have a look at git rebase. It is an awesome command to work with
patch series.

Thanks.

kkloginid:
changes were checked into main tree after git rebase so closing this request.

PR #143, opened 2016-03-10
Android: only redefine statvfs if it is not already defined

jbkempf:
It is defined with recent platforms version 21 and more recent


sahlberg:
Merged. Thanks!

PR #146, opened 2016-05-13
Pagecache v2

plieven:
Hi Ronnie,

as promised here is a rebased and slightly optimized version of my pagecache implementation from July/2015. Main changes include support for setting a cache ttl and ability to invalidate the cache from external.

Changes for qemu can be found here:
https://github.com/plieven/qemu/commits/nfs_pagecache_v2

I will submit them if this version of the patch is accepted.

Peter

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -78,23 +78,14 @@ fi
 
 AC_SUBST(MAYBE_EXAMPLES)
 
-
-WERROR_CFLAGS=""
-AC_MSG_CHECKING(whether to set -Werror)
-AC_ARG_ENABLE(werror, [  --disable-werror  do not treat warnings as errors during build],
-[ case "${enableval}" in
-  no)
-    AC_MSG_RESULT(no)
-    ;;
-  *)
-    AC_MSG_RESULT(yes)
-    WERROR_CFLAGS="-Werror"
-    ;;
-  esac ],
-)
+AC_ARG_ENABLE([werror], [AS_HELP_STRING([--disable-werror],
+              [Disables building with -Werror by default])])
 
 if test "$ac_cv_prog_gcc" = yes; then
-   WARN_CFLAGS="-Wall -Wshadow -Wno-write-strings -Wstrict-prototypes -Wpointer-arith -Wcast-align -Wno-strict-aliasing $WERROR_CFLAGS"
+   WARN_CFLAGS="-Wall -Wshadow -Wno-write-strings -Wstrict-prototypes -Wpointer-arith -Wcast-align -Wno-strict-aliasing"
+   if test "x$enable_werror" != "xno"; then
+       WARN_CFLAGS="$WARN_CFLAGS -Werror"
+   fi
 fi
 AC_SUBST(WARN_CFLAGS)
 
--- a/include/libnfs-private.h
+++ b/include/libnfs-private.h
@@ -32,6 +32,13 @@
 extern "C" {
 #endif
 
+#ifndef MIN
+#define MIN(a, b) (((a) < (b)) ? (a) : (b))
+#endif
+#ifndef MAX
+#define MAX(a, b) (((a) > (b)) ? (a) : (b))
+#endif
+
 #if !defined(HAVE_SOCKADDR_STORAGE) && !defined(WIN32)
 /*
  * RFC 2553: protocol-independent placeholder for socket addresses
@@ -119,6 +126,8 @@ struct rpc_context {
 	int uid;
 	int gid;
 	uint32_t readahead;
+	uint32_t pagecache;
+	uint32_t pagecache_ttl;
 	int debug;
 	int timeout;
 };
@@ -187,6 +196,8 @@ void rpc_unset_autoreconnect(struct rpc_context *rpc);
 void rpc_set_tcp_syncnt(struct rpc_context *rpc, int v);
 void rpc_set_uid(struct rpc_context *rpc, int uid);
 void rpc_set_gid(struct rpc_context *rpc, int gid);
+void rpc_set_pagecache(struct rpc_context *rpc, uint32_t v);
+void rpc_set_pagecache_ttl(struct rpc_context *rpc, uint32_t v);
 void rpc_set_readahead(struct rpc_context *rpc, uint32_t v);
 void rpc_set_debug(struct rpc_context *rpc, int level);
 void rpc_set_timeout(struct rpc_context *rpc, int timeout);
--- a/include/nfsc/libnfs.h
+++ b/include/nfsc/libnfs.h
@@ -37,8 +37,10 @@ extern "C" {
 #endif
 
 #define LIBNFS_FEATURE_READAHEAD
+#define LIBNFS_FEATURE_PAGECACHE
 #define LIBNFS_FEATURE_DEBUG
 #define NFS_BLKSIZE 4096
+#define NFS_PAGECACHE_DEFAULT_TTL 5
 
 struct nfs_context;
 struct rpc_context;
@@ -219,9 +221,16 @@ EXTERN uint64_t nfs_get_writemax(struct nfs_context *nfs);
 EXTERN void nfs_set_tcp_syncnt(struct nfs_context *nfs, int v);
 EXTERN void nfs_set_uid(struct nfs_context *nfs, int uid);
 EXTERN void nfs_set_gid(struct nfs_context *nfs, int gid);
+EXTERN void nfs_set_pagecache(struct nfs_context *nfs, uint32_t v);
+EXTERN void nfs_set_pagecache_ttl(struct nfs_context *nfs, uint32_t v);
 EXTERN void nfs_set_readahead(struct nfs_context *nfs, uint32_t v);
 EXTERN void nfs_set_debug(struct nfs_context *nfs, int level);
 EXTERN void nfs_set_dircache(struct nfs_context *nfs, int enabled);
+
+/*
+ *  Invalidate the pagecache
+ */
+EXTERN void nfs_pagecache_invalidate(struct nfs_context *nfs, struct nfsfh *nfsfh);
 
 /*
  * Sets timeout in milliseconds. A negative value means infinite timeout.
--- a/lib/init.c
+++ b/lib/init.c
@@ -80,6 +80,7 @@ struct rpc_context *rpc_init_context(void)
 	salt += 0x01000000;
 	rpc->fd = -1;
 	rpc->tcp_syncnt = RPC_PARAM_UNDEFINED;
+	rpc->pagecache_ttl = NFS_PAGECACHE_DEFAULT_TTL;
 #if defined(WIN32) || defined(ANDROID)
 	rpc->uid = 65534;
 	rpc->gid = 65534;
@@ -97,11 +98,44 @@ struct rpc_context *rpc_init_context(void)
 	return rpc;
 }
 
+uint32_t static round_to_power_of_two(uint32_t x) {
+	uint32_t power = 1;
+	while (power < x) {
+		power <<= 1;
+	}
+	return power;
+}
+
+void rpc_set_pagecache(struct rpc_context *rpc, uint32_t v)
+{
+	assert(rpc->magic == RPC_CONTEXT_MAGIC);
+	v = MAX(rpc->pagecache, round_to_power_of_two(v));
+	RPC_LOG(rpc, 2, "pagecache set to %d pages of size %d", v, NFS_BLKSIZE);
+	rpc->pagecache = v;
+}
+
+void rpc_set_pagecache_ttl(struct rpc_context *rpc, uint32_t v) {
+	if (v) {
+		RPC_LOG(rpc, 2, "set pagecache ttl to %d seconds\n", v);
+	} else {
+		RPC_LOG(rpc, 2, "set pagecache ttl to infinite");
+	}
+	rpc->pagecache_ttl = v;
+}
+
 void rpc_set_readahead(struct rpc_context *rpc, uint32_t v)
 {
 	assert(rpc->magic == RPC_CONTEXT_MAGIC);
-
+	if (v) {
+		v = MAX(NFS_BLKSIZE, round_to_power_of_two(v));
+	}
+	RPC_LOG(rpc, 2, "readahead set to %d byte", v);
 	rpc->readahead = v;
+	if (v) {
+		/* current pagecache implementation needs a pagecache bigger
+		 * than the readahead size to avoid collisions */
+		rpc_set_pagecache(rpc, (2 * v) / NFS_BLKSIZE);
+	}
 }
 
 void rpc_set_debug(struct rpc_context *rpc, int level)
--- a/lib/libnfs-win32.def
+++ b/lib/libnfs-win32.def
@@ -53,6 +53,7 @@ nfs_parse_url_full
 nfs_parse_url_dir
 nfs_parse_url_incomplete
 nfs_destroy_url
+nfs_pagecache_invalidate
 nfs_pread
 nfs_pread_async
 nfs_pwrite
@@ -71,6 +72,8 @@ nfs_set_auth
 nfs_set_debug
 nfs_set_dircache
 nfs_set_gid
+nfs_set_pagecache
+nfs_set_pagecache_ttl
 nfs_set_readahead
 nfs_set_tcp_syncnt
 nfs_set_timeout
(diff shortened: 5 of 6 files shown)


MilhouseVH:
Enabling `-Werror` by default results in the following build failure:
```
/home/neil/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-8.0-devel/libnfs-1.10.0/lib/libnfs.c:944:22: error: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'uint64_t {aka long long unsigned int}' [-Werror=format=]
   rpc_set_error(rpc, "server max rsize of %lu is greater than libnfs supported %d bytes",
                      ^
/home/neil/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-8.0-devel/libnfs-1.10.0/lib/libnfs.c:952:22: error: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'uint64_t {aka long long unsigned int}' [-Werror=format=]
   rpc_set_error(rpc, "server max wsize of %lu is greater than libnfs supported %d bytes",
                      ^
cc1: all warnings being treated as errors
```
Sure, I can disable `-Werror` again but shouldn't libnfs build without it if you're making it the default? Perhaps moving the `-Werror` change to a separate PR would be a better idea as it doesn't seem particularly relevant to this Pagecache change, and then any remaining build errors could be discussed in isolation.

sahlberg:
Why do you need this as a public function ?
nfs_pagecache_invalidate

The patch looks sane. Let me test it a bit with kodi and I will merge in a day or so.

Thanks!

plieven:
@sahlberg look at https://github.com/plieven/qemu/commit/2a4ed128ce6230882173e5236ba1e1ccf5f97fbe
function nfs_invalidate_cache ;-)

plieven:
@MilhouseVH In fact -Werror should be default since commit 84607f4, but it got broken in 617cfba. Is the above the only error you see? This should be easy to fix.

plieven:
@MilhouseVH please try again

MilhouseVH:
Unfortunately the next warning-as-error is:
```
/home/neil/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-8.0-devel/libnfs-1.10.0/lib/libnfs-sync.c: In function 'send_nfsd_probes':
/home/neil/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-8.0-devel/libnfs-1.10.0/lib/libnfs-sync.c:1713:9: error: cast increases required alignment of target type [-Werror=cast-align]
   ifr = (struct ifreq *)ptr;
         ^
cc1: all warnings being treated as errors
```
Perhaps a separate PR to fix these remaining 32-bit `-werror` issues?

sahlberg:
I will set up a 32bit machine and track this down. I don't have time right
today but should be able to do it in a few days.

plieven:
Hi Ronnie,

have you had a chance to test this with Kodi? Any speed improvements?

Thanks,
Peter

plieven:
@MilhouseVH I tried to compile on 32bit linux and it works. So the problem seems to be architecture independent. This means in return that nfs discovery would have never worked on a RPi2.
Can you try if the following patch works for you?

diff --git a/lib/libnfs-sync.c b/lib/libnfs-sync.c
index eba6121..2adc8d6 100644
--- a/lib/libnfs-sync.c
+++ b/lib/libnfs-sync.c
@@ -1707,39 +1707,39 @@ static int send_nfsd_probes(struct rpc_context *rpc, struct ifconf *ifc, struct
        assert(rpc->magic == RPC_CONTEXT_MAGIC);
 
        for (ptr =(char *)(ifc->ifc_buf); ptr < (char *)(ifc->ifc_buf) + ifc->ifc_len; ) {
-               struct ifreq *ifr;
+               struct ifreq ifr;
                char bcdd[16];
 
-               ifr = (struct ifreq *)ptr;
+               memcpy(&ifr, ptr, sizeof(struct ifreq));
 #ifdef HAVE_SOCKADDR_LEN
-               if (ifr->ifr_addr.sa_len > sizeof(struct sockaddr)) {
-                       ptr += sizeof(ifr->ifr_name) + ifr->ifr_addr.sa_len;
+               if (ifr.ifr_addr.sa_len > sizeof(struct sockaddr)) {
+                       ptr += sizeof(ifr.ifr_name) + ifr.ifr_addr.sa_len;
                } else {
-                       ptr += sizeof(ifr->ifr_name) + sizeof(struct sockaddr);
+                       ptr += sizeof(ifr.ifr_name) + sizeof(struct sockaddr);
                }
 #else
                ptr += sizeof(struct ifreq);
 #endif
 
-               if (ifr->ifr_addr.sa_family != AF_INET) {
+               if (ifr.ifr_addr.sa_family != AF_INET) {
                        continue;
                }
-               if (ioctl(rpc_get_fd(rpc), SIOCGIFFLAGS, ifr) < 0) {
+               if (ioctl(rpc_get_fd(rpc), SIOCGIFFLAGS, &ifr) < 0) {
                        return -1;
                }
-               if (!(ifr->ifr_flags & IFF_UP)) {
+               if (!(ifr.ifr_flags & IFF_UP)) {
                        continue;
                }
-               if (ifr->ifr_flags & IFF_LOOPBACK) {
+               if (ifr.ifr_flags & IFF_LOOPBACK) {
                        continue;
                }
-               if (!(ifr->ifr_flags & IFF_BROADCAST)) {
+               if (!(ifr.ifr_flags & IFF_BROADCAST)) {
                        continue;
                }
-               if (ioctl(rpc_get_fd(rpc), SIOCGIFBRDADDR, ifr) < 0) {
+               if (ioctl(rpc_get_fd(rpc), SIOCGIFBRDADDR, &ifr) < 0) {
                        continue;
                }
-               if (getnameinfo(&ifr->ifr_broadaddr, sizeof(struct sockaddr_in), &bcdd[0], sizeof(bcdd), NULL, 0, NI_NUMERICHOST) < 0) {
+               if (getnameinfo(&ifr.ifr_broadaddr, sizeof(struct sockaddr_in), &bcdd[0], sizeof(bcdd), NULL, 0, NI_NUMERICHOST) < 0) {
                        continue;
                }
                if (rpc_set_udp_destination(rpc, bcdd, 111, 1) < 0) {

plieven:
`diff --git a/lib/libnfs-sync.c b/lib/libnfs-sync.c
index eba6121..2adc8d6 100644
--- a/lib/libnfs-sync.c
+++ b/lib/libnfs-sync.c
@@ -1707,39 +1707,39 @@ static int send_nfsd_probes(struct rpc_context *rpc, struct ifconf *ifc, struct
        assert(rpc->magic == RPC_CONTEXT_MAGIC);
 
        for (ptr =(char *)(ifc->ifc_buf); ptr < (char *)(ifc->ifc_buf) + ifc->ifc_len; ) {
-               struct ifreq *ifr;
+               struct ifreq ifr;
                char bcdd[16];
 
-               ifr = (struct ifreq *)ptr;
+               memcpy(&ifr, ptr, sizeof(struct ifreq));
 #ifdef HAVE_SOCKADDR_LEN
-               if (ifr->ifr_addr.sa_len > sizeof(struct sockaddr)) {
-                       ptr += sizeof(ifr->ifr_name) + ifr->ifr_addr.sa_len;
+               if (ifr.ifr_addr.sa_len > sizeof(struct sockaddr)) {
+                       ptr += sizeof(ifr.ifr_name) + ifr.ifr_addr.sa_len;
                } else {
-                       ptr += sizeof(ifr->ifr_name) + sizeof(struct sockaddr);
+                       ptr += sizeof(ifr.ifr_name) + sizeof(struct sockaddr);
                }
 #else
                ptr += sizeof(struct ifreq);
 #endif
 
-               if (ifr->ifr_addr.sa_family != AF_INET) {
+               if (ifr.ifr_addr.sa_family != AF_INET) {
                        continue;
                }
-               if (ioctl(rpc_get_fd(rpc), SIOCGIFFLAGS, ifr) < 0) {
+               if (ioctl(rpc_get_fd(rpc), SIOCGIFFLAGS, &ifr) < 0) {
                        return -1;
                }
-               if (!(ifr->ifr_flags & IFF_UP)) {
+               if (!(ifr.ifr_flags & IFF_UP)) {
                        continue;
                }
-               if (ifr->ifr_flags & IFF_LOOPBACK) {
+               if (ifr.ifr_flags & IFF_LOOPBACK) {
                        continue;
                }
-               if (!(ifr->ifr_flags & IFF_BROADCAST)) {
+               if (!(ifr.ifr_flags & IFF_BROADCAST)) {
                        continue;
                }
-               if (ioctl(rpc_get_fd(rpc), SIOCGIFBRDADDR, ifr) < 0) {
+               if (ioctl(rpc_get_fd(rpc), SIOCGIFBRDADDR, &ifr) < 0) {
                        continue;
                }
-               if (getnameinfo(&ifr->ifr_broadaddr, sizeof(struct sockaddr_in), &bcdd[0], sizeof(bcdd), NULL, 0, NI_NUMERICHOST) < 0) {
+               if (getnameinfo(&ifr.ifr_broadaddr, sizeof(struct sockaddr_in), &bcdd[0], sizeof(bcdd), NULL, 0, NI_NUMERICHOST) < 0) {
                        continue;
                }
                if (rpc_set_udp_destination(rpc, bcdd, 111, 1) < 0) {
`

MilhouseVH:
Thanks. This is the build log for libnfs (latest master) with https://github.com/plieven/libnfs/commit/b10ffed267b65247cb277c75fa4c6856539c7761 when building LibreELEC for RPi2: http://sprunge.us/PXNW - problem now is with libnfs-zdr.c.

plieven:
Uhh, this will get a little more complicated. Is there a possibility to cross-compile for RPI2 somewhere? Or can I manually trigger a rebuild of your system with the latest checkout of a specific branch?

MilhouseVH:
You could try building LibreELEC (PROJECT=RPi2 ARCH=arm) for yourself, but it's a patch-based build system which makes developing using LibreELEC rather difficult/tedious/frustrating.

However if you want to continue posting patches (here, or on a new PR?) I'll be happily to build them and post the results until we get a clean RPi build.

If you create a new PR and keep it updated with your patches then I can re-build LibreELEC by applying the latest commits from the PR (ping me in the PR when it's updated so that I know when to re-build).

sahlberg:
That is a very odd error.
It complains that the cast will expand from (char *) to (uint32_t *) type,
which for legacy reasons should always be allowed.
(or a LOT of old sw will complain)

A workaround is to not cast a (char *) to something larger but using memcpy
between the different types.


The attached patch, very lightly tested with valgrind and manual testing)
uses memcpy to copy the data instead of casting the pointer and assigning.


Can you test if you get these warning with this too ?

MilhouseVH:
Doesn't look like the patch made it.

sahlberg:
Ok, I pushed it to a temp branch instead.

Try the patch 4ae9c77cb63a327344497db7a2f2efc277f51cb7 in my branch
"alignment"

MilhouseVH:
Thanks, that seems to OK. Now it fails a little later on: http://sprunge.us/YGJN

plieven:
@sahlberg we should find a different solution than memcpy. I can look into this tomorrow. I found out that ~90% of libnfs cpu cycles are burnt with memcpy. I am currently working on a branch to eliminate some of the memcmpies. I have an idea for a different solution for the (u_int32_t)* cast, but no time till tomorrow. Looking at the CPU cycles the (u_int32_t)* cast also seems not so cheap on x86_64.

sahlberg:
It is not surprising that most time is spent in memcpy. NFS/RPC is
unfortunately very difficult to do zero -copy in userspace :-( due to the
fact that the headers have wildly varying length so it is never easy to
know where in the bytestream the payload for read/write.   This means that
all read/write payload will need to go through copy via at least one buffer
:-(

The problem Milhouse is facing is not I think related to platform but
rather compiler.  His compiler is generating warnings for the common
pattern of casting a (char *) buf pointer to something bigger than a char.
Perhaps the best solution here is just simply disable these warnings on his
compiler.

Milhouse, what is the exact compiler you use?

MilhouseVH:
It is gcc-5.3.0.

I'm building on Ubuntu 16.04 x86_64 where gcc-5.3.0 is built as part of the LibreELEC toolchain, and cross-compiling for ARMv7.

MilhouseVH:
Do either of you have access to a Raspberry Pi running Raspbian?

I just tried to build libnfs master on the Raspberry Pi (Pi3,  gcc 4.9.2, latest "Jessie") so no cross-compiling, and experienced the following build failure: http://pastebin.com/bKSKcirj

With `--disable-werror` it succeeds: http://pastebin.com/kHZ9U9rm - perhaps the various compiler warnings may prove useful in tracking down the remaining issues?

If you have a Pi, building on the Pi may help identify the failures quicker than setting up your own cross-compile environment.

However if you are willing to try using LibreELEC as a cross-compile environment for RPi2 it's fairly straightforward:
```
#Ubuntu 15.04 lacks the following dependencies, so add them:
sudo apt-get install curl gawk gperf xsltproc openjdk-8-jre-headless libncurses5-dev texinfo ccache u-boot-tools

mkdir projects
cd projects
git clone git://github.com/LibreELEC/LibreELEC.tv.git
wget -q www.nmacleod.com/public/oebuild/oebuild.tar.gz -O - | tar xzf -
mv build.conf.sample build.conf
./packages/make-libnfs.sh
cd LibreELEC.tv
PROJECT=RPi2 ARCH=arm scripts/build libnfs
```
And - once the toolchain has been built, which may take a while (an hour or so on an FX-8350) - you'll be building the latest libnfs master, which should fail to build due to the `Werror` issues.

Any libnfs patches will need to be copied into LibreELEC.tv/projects/RPi2/patches/libnfs, then:
```
PROJECT=RPi2 ARCH=arm scripts/clean libnfs
PROJECT=RPi2 ARCH=arm scripts/build libnfs
```
to rebuild libnfs.

plieven:
@sahlberg I already was able to eliminate some memcpy's. Regarding the uint32_t* cast, I have a solution that works. I will compare the costs with the cast tomorrow.

@MilhouseVH I will fix the constant error you observe on the RPi.
Thanks also for the howto for the toolchain I will try that out.

sahlberg:
Ok, there were this, and one other place, where still cast a char * to
<something else> *

I have changed these to cast via void *.
void * is special, it allows casting from/to any other type.

Before void * existed (early 80s?) it was common to use char * when casting.
When void * was added as the proper way to have a pointer type you could
case anything to/from
c implementors still allowed this same property for char *  since otherwise
enourmous amount
of software would stop compiling.

Looks like now, almost 30 years later, they have finally bitten the bullet
and made this a warning or error (with -Werror)
That is sure some very long convervative time for a language to phase out a
language feature :-) Wow.


Try this from the tip of my alignment branch.
1575ed244b3992bdab3dd23a911fefdd2b6ee3e2

sahlberg:
I used to have a raspberry pi model a in my build farm. Sadly that machine
died a few years back :-(

I will try to set up a raspbian image under qemu.

MilhouseVH:
Thanks @sahlberg.

Latest build failure on Ubuntu x86 cross-compiling to ARMv7 (RPi2): http://sprunge.us/JaNW

I'm building libnfs master (654243473adc3f3f2b2e49f09741f7d84261158f) plus the following two patches:

https://github.com/plieven/libnfs/commit/b10ffed267b65247cb277c75fa4c6856539c7761
https://github.com/sahlberg/libnfs/commit/1575ed244b3992bdab3dd23a911fefdd2b6ee3e2

sahlberg:
One more try. Sorry for the mess.

This got to work, goddamit
a35a274fa812c4f3b5bc554eaedaad93f124a6ad

Or else it will have to wait a few days until I get the rpi emulator up and
running.

MilhouseVH:
We have a successful RPi2 build! :)

I'll build LE x86_64 with the same patches, and run tests in Kodi (both RPi2 and x86_64).

If there's no obvious problems I'll include them in nightly test builds (x86 and RPi) and will report back in a few days if there are any reports of problems.

MilhouseVH:
I've now built LibreELEC for RPi2 and x86_64 with these latest patches and they both build OK, and both play movies/music OK in Kodi. Seeking is also working fine, and there are no obvious regressions, so it's looking good.

I'll include both patches in the nightly test builds for wider testing, starting Thursday night.

Many thanks to you both.

plieven:
please have a look at #150 

@MilhouseVH can you please also check building natively on the RPi2. Namely the constant error.

PR #150, opened 2016-06-02
For upstream

plieven:


lib/libnfs-zdr.c
```diff
@@ -54,6 +54,14 @@ struct zdr_mem {
 
 struct opaque_auth _null_auth;
 
+inline uint32_t libnfs_buf2hl(char *buf) {
+	return ntohl(*(uint32_t *)(void *)buf);
```
sahlberg: Could probably change the buf argument to b void *buf and avoid the explicit cast.
But doesn't matter.

I can merge as is once Milhouse verifies it works (or once I set up a rpi emulator)

plieven:
@sahlberg good comment, changed it.

@MilhouseVH I fail to setup the libreelec toolchain. can you advise:

root@libreleec-toolchain:/usr/src/projects# ./packages/make-libnfs.sh
Loading profile: pi2 [15:12:27] (./supermk.sh libnfs.git ./tools/mkpkg_libnfs  network/libnfs/package.mk)
ERROR: Failure querying /usr/src/projects/LibreELEC.tv for "CCACHE_DIR", check PROJECT/ARCH [RPi2/arm] settings
ERROR: Failure querying /usr/src/projects/LibreELEC.tv for "DISTRO", check PROJECT/ARCH [RPi2/arm] settings
Failure obtaining DISTRO

MilhouseVH:
Weird, cross-compiling for ARMv7 (RPi2) on Ubuntu (gcc 5.3.0) I'm getting this build failure: http://sprunge.us/QQAV

MilhouseVH:
Apparently it may require ` --std=gnu89` in CFLAGS (as one solution) but maybe there's a better way which doesn't require specific flags? gcc 5.x has changed default std from 89 to gnull.

MilhouseVH:
LE x86 build has also failed (same reason as RPi).

About to head out for the evening, if you have any changes I can test them tomorrow (maybe overnight).

plieven:
please try to build without 5d0f0b7

MilhouseVH:
> please try to build without 5d0f0b7

Yes, that cross-compiles OK with gcc-5.3.0, and Kodi also runs fine (on both RPi2 and x86).

plieven:
I removed that patch from the pull request it was only cosmetic. @sahlberg I think you can go ahead.

plieven:
@MilhouseVH regarding the cross compiler tool chain. I copied the sample conf and nevertheless get the error. can you please verify the commands. Thanks.

MilhouseVH:
@plieven I've just tested [the instructions](https://github.com/sahlberg/libnfs/pull/146#issuecomment-223077151) on a fresh Ubuntu 16.04 VM (VirtualBox) and the instructions are working as expected (the only issue being `interdiff` which isn't available in 16.04 - fortunately this isn't an issue as you don't need it to build libnfs - and I also needed to install `git`, as it was a fresh Ubuntu).

So really not sure what the problem is here. Although, when you posted [the error](https://github.com/sahlberg/libnfs/pull/150#issuecomment-223286957), you seemed to be running as `root` in `/usr/src` which could be part of the problem... I'd suggest executing the instructions as a normal user in `/home` etc., that's where I run the instructions without any issue.

PR #151, opened 2016-06-07
nfs_rewinddir: introduce additional nfs_rewinddir()

benoit-canet:
Introduce the nfs_rewinddir() utility function used
to cancel previous nfs_readdir() side effects.

Signed-off-by: Benoît Canet <benoit@scylladb.com>

### DIFF
--- a/include/nfsc/libnfs.h
+++ b/include/nfsc/libnfs.h
@@ -949,6 +949,14 @@ struct nfsdirent  {
 EXTERN struct nfsdirent *nfs_readdir(struct nfs_context *nfs, struct nfsdir *nfsdir);
 
 
+/*
+ * REWINDDIR()
+ */
+/*
+ * nfs_rewinddir() cancel all previous nfs_readdir() side effects
+ */
+EXTERN void nfs_rewinddir(struct nfs_context *nfs, struct nfsdir *nfsdir);
+
 
 /*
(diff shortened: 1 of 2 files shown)


nyh:
Sounds to me like rewinddir() might not be enough - if you want to be able to allow the user to seek to arbitrary locations in the directory (see seekdir(3) and telldir(3)), you need methods like nfs_seekdir() and nfs_telldir() - and rewinddir will be a special case, nfs_seekdir(0).

benoit-canet:
@sahlberg Do you have any comments on this idea ? We hit this in our OSv libnfs integration.

sahlberg:
Merged, thanks.

It would be nice to have seekdir/telldir and have rewinddir just be a special case, but as these functions are rarely used I don't plan to implement them until there is a real need.

sahlberg:
On Tue, Jun 7, 2016 at 3:35 AM, Benoît Canet <notifications@github.com>
wrote:

> @sahlberg <https://github.com/sahlberg> Do you have any comments on this
> idea ? We hit this in our OSv libnfs integration.
>

It is fine as is for now. It is a trivial function and I merged it.

If when the need for seekdir/telldir arise  we can trivially convert
rewinddir to become a special case of seekdir.

Thanks for your contribution !

PR #155, opened 2016-07-29
Added the ability to bind to a specific interface

doktorstick:
Added the ability to bind to a specific interface with the `nfs_set_interface` and `rpc_set_interface` APIs, or via the NFS URL `if=<interface>` parameter. This feature requires `root` permissions.

NOTE: This has only been compiled and tested on Ubuntu 14.04. It's unlikely that it'll work on other platforms without modification, particularly around the inclusion of `<net/if.h>` and `IFNAMSIZ` define in `libnfs-private.h`.


sahlberg:
Thanks for this patch!

I merged this patch after adding minor changes to make it conditional upon SO_BINDTODEVICE support and the configure magic to check for it.

PR #158, opened 2016-12-22
For upstream

plieven:
Hi Ronnie,

libnfs suffers from the same issue in rpc_read_from_socket. This patch changes the behaviour similar to libnfs. However, it only reads what is available when we enter rpc_read_from_socket to avoid starvation in rpc_read_from_socket. I also found that we nowhere check if the incoming PDU is too small. I added a check for 0 bytes, but I have no idea what the mimum PDU size really is. Maybe you have an idea.

Peter


sahlberg:
Merged, thanks.

PR #160, opened 2016-12-27
Return ESTALE for NFS3ERR_STALE, not EIO.

kevin-vigor:
ESTALE is retryable in many circumstances, EIO is not.


sahlberg:
Merged, thanks

PR #162, opened 2017-01-05
Revert "socket: batch pdu read in function rpc_read_from_socket"

plieven:
The ioctl version breaks Qemu. I will post an updated once we found
a good solution in libiscsi and then adapt it to libnfs.

This reverts commit 003b3c7ce28729aad44deff442186f4a70011745.


sahlberg:
okidoki

sahlberg:
I really wonder why QEMU has a problem with ioctl(FIONREAD)

plieven:
me, too. i see this I thread spun message and the i/o thread deadlocks.

Von meinem iPhone gesendet

PR #165, opened 2017-01-24
Fix Windows jukebox issue with NetApp on Windows clients.

keeely:


sahlberg:
There is a public api now that allows to set the uid/gid directly from the application.

PR #166, opened 2017-02-03
Allow passing void* buffers, and make write buffers const

MaxKellermann:
Enables callers to pass any opaque data chunk without having to cast
it explicitly.

A write never modifies the source buffer, and thus the pointer should
be const.

Signed-off-by: Max Kellermann <max.kellermann@gmail.com>


sahlberg:
Merged, thanks.

PR #169, opened 2017-02-08
init: remove pdu from list before invoking callback

MaxKellermann:
If the callback does anything fishy that modifies the linked list,
libnfs may crash after returning.  So doing any pending list removals
before invoking the callbacks is safer.


sahlberg:
Merged, thanks

PR #173, opened 2017-02-26
socket: Limit reconnect retries to 10

rosslagerwall:
Limit the number of retries when autoreconnecting (to an arbitrary 10)
and return an error to the application if this limit is reached.
Without this, libnfs retries indefinitely and consumes 100% CPU.

See also: https://bugzilla.gnome.org/show_bug.cgi?id=762544

Signed-off-by: Ross Lagerwall <rosslagerwall@gmail.com>

### DIFF
--- a/include/libnfs-private.h
+++ b/include/libnfs-private.h
@@ -133,6 +133,7 @@ struct rpc_context {
 	/* track the address we connect to so we can auto-reconnect on session failure */
 	struct sockaddr_storage s;
 	int auto_reconnect;
+	int auto_reconnect_retries;
 
 	/* fragment reassembly */
 	struct rpc_fragment *fragments;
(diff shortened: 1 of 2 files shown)


sahlberg:
Thanks. Merged.

We may want to ensure that this loop will also be time-based so that we dont end up spinning these 10 iterations in just a few ms.

We should also make this a tunable so we can tweak this at runtime from applications.
Adding a rpc_set_reconnect_count() etc that would default to some sane value.

But this is a start.

sahlberg:
Merged, thanks!

Did you look into using the functions that search for/lists nfs servers as
well as listing the exports on a specific server  so that the experience
from nautilus would be similar to smb shares ?

rosslagerwall:
> Did you look into using the functions that search for/lists nfs servers as well as listing the exports on a specific server so that the experience from nautilus would be similar to smb shares ?

Not yet, sorry...

PR #174, opened 2017-02-26
Here are some changes to resolve some conflicts with MinGW and fuse_dokan

Daniel-Abrecht:
When I tried do cross-compile this library with x86_64-w64-mingw32-gcc and i686-w64-mingw32-gcc, i got some conflicts between mingw and win32_compat.h. Also, when I tried to cross-compile fuse-nfs and dokan_fuse, I got some conflicts between the version of struct statvfs provided by libnfs and dokan_fuse. This pull request resolves both problems.


sahlberg:
Merged, thanks!

PR #177, opened 2017-03-25
Cygwin should not use winsock2

billziss-gh:
As per the commit message:

> The configure.ac script was erroneously including -lws2_32,
> i.e. the winsock2 library. Cygwin is a full POSIX environment
> and does not need the winsock2 functionality as it supplies its
> own POSIX compatible sockets.
>
> Including -lws2_32 caused getnameinfo to fail with error 10093
> (WSANOTINITIALISED). This usually means that the Winsock
> WSAStartup function has not been called. Calling WSAStartup is
> not necessary for Cygwin programs and in fact should be avoided.


sahlberg:
Merged. Thanks.

billziss-gh:
Thanks again!

PR #178, opened 2017-04-03
Make nfs_close_async call flush on the file handle before disposing of it.

kevin-vigor:


sahlberg:
Merged, thanks.

An optional enhancement would be to track when we have do (meta-)data changes to the filehandle
and only call fsync on close IFF there are any changes that need to be destaged.

As fsync needs to walk all the extents of the file I think we might be wasting some time/cpu if we flush
and there is nothing to destage. For example after close on a O_RDONLY file we don't need to waste time on fsync. (I think fsync could become somewhat expensive for very large files)


No need to do that change right now though. I can add that later.

pkerling:
@sahlberg We're seeing some issues with this in Kodi exactly for the case you mentioned (closing an `O_RDONLY` file). We seem to get `NFS3ERR_ROFS` from the server on `COMMIT` of such files (side-note: the export is read-only). Do you have any insight whether this is a violation of the spec on the server side or should the client not send `COMMIT` when there are no writes to a file like you suggested?

sahlberg:
The client shouldn't send a commit for that case. (though the commit is
harmless and the error should be ignored)

I have pushed a simple change to stop sending a commit on close if the file
has not been written to.
I aldready did this for NFSv4 but forgot to add the same logic to NFSv3.

Try current master.

Issue #187, opened 2017-05-01
Proposal for nfs_readlink() to return content without incurring ENAMETOOLONG


earlchew:
Allow a caller using the sync API to read the content of a symbolic link without incurring ENAMETOOLONG.

The NFS SYMLINK RPC (https://tools.ietf.org/html/rfc1813#page-61) returns the content of the symbolic link, but the current API requires that the result be copied into a buffer provided by the caller, leading to an ENAMETOOLONG error if that buffer is too small.

This proposal is to either change the current API, or provide an alternate API (eg nfs_readlink2) that will return the symbolic link content to the caller, and making it the responsibility of the caller to release the memory used to hold the link content.

Here is the current API:
```
/*
 * Sync readlink(<name>)
 * Function returns
 *      0 : The operation was successful.
 * -errno : The command failed.
 */
EXTERN int nfs_readlink(struct nfs_context *nfs, const char *path, char *buf, int bufsize);
```

Some options:
* Return a `char *`, and use `NULL` to indicate an error
* Continue to use `int` for returning status, but require the caller to pass `char **bufptr` which the function will use to return a `char *`.
* Continue to use `int` for returning status, but require the caller to pass `char **bufptr` and `int bufsize` which the caller can use to either provide a buffer and specify its length, or provide `*bufptr == NULL` to indicate that the function should return a buffer to the caller. Maybe this option is really only useful if it's desirable to rework the existing `nfs_readlink`.

sahlberg:
I think that change makes sense.
I am planning a new release for next week to announce all the major new things such as nfsv4 and rpc server support so this is a good time to bump the api version and change some function signatures.

Thanks for the suggestion,  I will try to find time to implement "Return a char *, and use NULL to indicate an error" in hte next few days so that it will make it into the next release.

earlchew:
Do you have a preferred approach? I can send you a pull request for this too.

sahlberg:
Change the current API and "Return a char *, and use NULL to indicate an
error"

Should be fine. This will break the API but that is ok because I plan to
bump the library version anyway when I release the next version of libnfs.

sahlberg:
Sorry,   please create a new function nfs_readlink2() that has the new behaviour.  I can flag the old command as deprecated in the documentation/libnfs.h

earlchew:
[...]

PR #194, opened 2017-05-05
nfs_readlink2: Return symlink without incurring ENAMETOOLONG

earlchew:


earlchew:
https://github.com/sahlberg/libnfs/issues/187

Issue #200, opened 2017-05-09
Leak in nfs_mount_8_cb() when called as a consequence of nfs_destroy_context()


earlchew:
A very interesting case. A call to `nfs_destroy_context()` causes PDUs to be dequeued with `RPC_STATUS_CANCEL`, but `nfs_mount_8_cb` attempts to reconnect:

```
finished:
         ...
        rpc_disconnect(rpc, "normal disconnect");

        if (rpc_connect_program_async(nfs->rpc, nfs->server, NFS_PROGRAM, NFS_V3, nfs_mount_9_cb, data) != 0) {
                data->cb(-ENOMEM, nfs, command_data, data->private_data);
                free_nfs_cb_data(data);
                return;
        }
```

```
==21843== 238 (32 direct, 206 indirect) bytes in 1 blocks are definitely lost in loss record 608 of 1,534
==21843==    at 0x4C2CC4B: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==21843==    by 0x916969F: rpc_connect_program_async (libnfs.c:796)
==21843==    by 0x9169DDF: nfs_mount_8_cb (libnfs.c:1090)
==21843==    by 0x9161C32: rpc_destroy_context (init.c:310)
==21843==    by 0x9168FAA: nfs_destroy_context (libnfs.c:540)
==21843==    by 0x8F3B740: ??? (in /usr/lib/python2.7/dist-packages/libnfs/_libnfs.so)
==21843==    by 0x49D189: call_function.9501 (ceval.c:4025)
==21843==    by 0x48BDDB: PyEval_EvalFrameEx.constprop.14 (ceval.c:2669)
==21843==    by 0x49D58C: call_function.9501 (ceval.c:4111)
==21843==    by 0x48BDDB: PyEval_EvalFrameEx.constprop.14 (ceval.c:2669)
==21843==    by 0x49C617: PyEval_EvalCodeEx (ceval.c:3257)
==21843==    by 0x4E6A0F: function_call (funcobject.c:526)
```

earlchew:
@sahlberg Is this the right approach?

```
diff --git a/lib/libnfs.c b/lib/libnfs.c
index af51a90..ae8fea2 100755
--- a/lib/libnfs.c
+++ b/lib/libnfs.c
@@ -1067,6 +1067,12 @@ finished:

        rpc_disconnect(rpc, "normal disconnect");

+        if (status == RPC_STATUS_CANCEL) {
[...]

PR #204, opened 2017-05-09
Provide error return from nfs_mount_8_cb()

earlchew:


earlchew:
https://github.com/sahlberg/libnfs/issues/200

sahlberg:
Merged, thanks!

Issue #206, opened 2017-05-10
Check for insane rtmax and wtmax values


earlchew:
Commit a523f843530d05b40353244709bdcf9fd5cd4d96 clamped readsize and writesize. Additionally, for safety do not proceed if either are zero because subsequent algorithms assume non-zero values.

PR #207, opened 2017-05-10
Return an error if either readmax or writemax are zero

earlchew:


earlchew:
https://github.com/sahlberg/libnfs/issues/206

sahlberg:
This patch does not merge cleanly. Please re-sync with upstream master.

I think I already did these changes after ganesha (which reports huge maximums) had issues with libnfs.

sahlberg:
Please re-sync with master and add the check for 0.
Or instead of  checkng for >0    you can do a check for >=8192  as all v3 servers must support at least this much.

Issue #212, opened 2017-05-10
Provide a way to disable autoreconnect


earlchew:
As a application user, I want to be able to disable autoreconnect so that I can reconnect my NFS client to another IP address in my storage cluster. With autoreconnect enabled, the NFS client tries to reconnect to an IP address that it no longer available.

sahlberg:
I see.
You basically want to have libnfs fail fast so that you can reconnect to a
different server ip instead of having to wait for the reconnection logic to
try and fail 10 consecutive tcp connection attempts (which may be a long
time).
That is a far use-case but I think we need more than this patch.

For the current use-case where we just try to reconnect back to the same
server, we still need to do some additional work in order to be able to
re-issue the PDUs that are in flight.
See rpc_reconnect_requeue().
We need to both re-set the number of bytes sent counter for the current pdu
as we might have just sent partial data to the server when the network
disconnect happen.
We also need to move any of the pdus in the different waitpdu buckets back
to the outqueue so that they will eventually be re-sent to the server once
the connection is recovered.
Otherwise we will drop all the pdus that were in flight during the network
session failure, which is not what we want.


So what I think you want instead is probably something almost like the
current rpc_reconnect_requeue() but switch the rpc sock_addr before you
invoke.


I think something like this :
1, Set reconnect attempts to 0.

2, once the connection fails you will notice this immediately by
rpc_service() returning an error.
3, in your app you now update the rpc structure to point to a different
sock_addr for the server you failed over to.
4, Set reconnect attempts to something sane, like 3 or so
5, once you are confident the new server is available, then invoke
rpc_reconnect_requeue() directly from you app
and hopefully it will now do all its goodness. Re-queue any pdus that are
in flight, and reconnect to the new target you swapped to.

6, once connected successfully, maybe set reconnect attempts back to 0
again?

sahlberg:
Maybe an even better, and a lot less intrusive change,  would be to add
code so you can register a "reconnect handler" to the rpc structure.
Then have rpc_reconnect_requeue() just invoke that handler before every try
to reconnect.

The flow would then be something like :

1, In your app, once the rpc/nfs context is created, register a function
pointer to your own reconnect handler :
     rpc_set_reconnect_handler(rpc, void *reconnect_handler(struct
rpc_context *rpc))


2, When the connection has failed, have rpc_reconnect_requeue() call out to
your handler
    if (rpc->rpc_reconnect_handler) {
          rpc->rpc_reconnect_handler(rpc);
[...]

PR #213, opened 2017-05-10
Allow autoreconnect to be disabled

earlchew:

### DIFF
--- a/include/nfsc/libnfs.h
+++ b/include/nfsc/libnfs.h
@@ -232,6 +232,7 @@ EXTERN void nfs_set_pagecache_ttl(struct nfs_context *nfs, uint32_t v);
 EXTERN void nfs_set_readahead(struct nfs_context *nfs, uint32_t v);
 EXTERN void nfs_set_debug(struct nfs_context *nfs, int level);
 EXTERN void nfs_set_dircache(struct nfs_context *nfs, int enabled);
+EXTERN void nfs_set_autoreconnect(struct nfs_context *nfs, int enabled);
 
 /*
  *  Invalidate the pagecache
--- a/lib/libnfs-win32.def
+++ b/lib/libnfs-win32.def
@@ -69,6 +69,7 @@ nfs_rmdir
 nfs_rmdir_async
 nfs_service
 nfs_set_auth
+nfs_set_autoreconnect
 nfs_set_debug
 nfs_set_dircache
 nfs_set_gid
--- a/lib/libnfs.c
+++ b/lib/libnfs.c
@@ -145,6 +145,7 @@ struct nfs_context {
        uint64_t writemax;
        char *cwd;
        int dircache_enabled;
+       int auto_reconnect_enabled;
        struct nfsdir *dircache;
        uint16_t	mask;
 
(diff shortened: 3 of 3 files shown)


earlchew:
https://github.com/sahlberg/libnfs/issues/212

sahlberg:
The patch looks fine, but I am a little curious about the use-case.
Traditionally, nfs clients default to try reconnecting forever and never giving up.

Libnfs will not retry forever, but retry reconnecting the socket 10 times. This would if the network/server is down be in the order of a few minutes, depending on how long connect() on the client takes to fail the connection attempt.
With this change libnfs will instead fail immediately and thus not be able to recover from brief network outages.

As this is controlled by the user and it defaults to not diable the 10 reconnect attempts, I am not against
adding this feature but just curious of the usecase.

If what you do want is just to have it fail much faster so that the application can try to recover maybe a better solution than the current "try 10 times, regardless how much wall-time that is" or your "dont try to recover at all" would be to be something timer based, i.e.   allow trying to reconnect until n seconds have passed and then give up.  N being user configurable.

earlchew:
> Traditionally, nfs clients default to try reconnecting forever and never giving up.

The use case that I have is an application that uses libnfs to connect to a distributed storage cluster where there are multiple end points, and all provide a consistent view onto the stored data. In this case, there is a desire to have better control over the time to cut over from one failed connection to another.

For this scenario, re-trying a failed connection indefinitely is undesirable, and failing as-soon-as-possible allows the retry policy to be entirely in the hands of the application. This is satisfactory for the use-case that I am interested in.

> If what you do want is just to have it fail much faster so that the application can try to recover maybe a better solution than the current "try 10 times, regardless how much wall-time that is" or your "dont try to recover at all" would be to be something timer based, i.e. allow trying to reconnect until n seconds have passed and then give up. N being user configurable.

If you would like to see this to be more configurable, I think there are two possibilities:
* Allow the number of retries to be configured: zero means to fail immediately, 10 would match the current behaviour.
* Allow the duration of retries to be configured: zero means fail immediately, 100s would mean retry for at least 100s. There is no way to configure this to be the same as the 10 reconnection attempts that the current implementation provides.

sahlberg:
See my comments in https://github.com/sahlberg/libnfs/issues/212
so that we can try doing the reconnect to a different server without dropping any PDUs in-flight and causing dataloss.


One additional thing you will need to do in order to get this to work is configuration management on your nfs server.
In order to be able to re-issue these nfs commands to the new server it is imperative that the filehandles MUST remain stable across the different servers. If they are not, the server will return -ESTALE which is a non-recoverable error :-(

If you have a cluster filesystem backing your clustered nfs servers, that cluter filesystem needs to guarantee that the inodes and filehandles are stable across the nodes in your storage cluster.
GPFS does guarantee this and can do failover between nodes completely transparently.
It also will migrate the ip addresses you access the nfs server through so you dont really need to have this "swap target socket address" support in libnfs.
I do not know how well other clusters filesystems handle this, but if they do not guarantee that filehandles are stable across the cluster, you will unfortunately have dataloss every time you do a failover due to -ESTALE (with libnfs as well as with kernel nfs client) :-(

Also make sure that you specify an explicit fsid=...  for every export in /etc/exports and have this in sync across all the server nodes so that fsid is stable as well.

sahlberg:
Merged, thanks.

PR #214, opened 2017-05-10
Return an error if either readmax or writemax are less than NFSMAXDATA2

earlchew:


earlchew:
https://github.com/sahlberg/libnfs/issues/206

sahlberg:
Nice, thanks!

Issue #215, opened 2017-05-10
Callbacks re-entered on rpc_disconnect() result in SIGSEGV


earlchew:
Here is an interesting case where `rpc_destroy_context()` invokes a callback which calls `rpc_disconnect()`, which then calls `rpc_error_all_pdus()` which re-enters the same callback:

```
#0  nfs_mount_8_cb (rpc=0x21a2cd0, status=1, command_data=0x7fba40f7b22f,
    private_data=0x21adf70) at libnfs.c:1066
#1  0x00007fba40f60462 in rpc_error_all_pdus (rpc=0x21a2cd0, error=
    0x7fba40f7b22f "normal disconnect") at init.c:250
#2  0x00007fba40f705cb in rpc_disconnect (rpc=0x21a2cd0, error=
    0x7fba40f7b22f "normal disconnect") at socket.c:622
#3  0x00007fba40f68949 in nfs_mount_8_cb (rpc=0x21a2cd0,
    status=<optimized out>, command_data=0x0, private_data=0x21adf70)
    at libnfs.c:1121
#4  0x00007fba40f606f1 in rpc_destroy_context (rpc=0x21a2cd0) at init.c:319
#5  0x00007fba40f67acb in nfs_destroy_context (nfs=0x21a7250) at libnfs.c:544
#6  0x00007fba41193741 in ?? ()
   from /usr/lib/python2.7/dist-packages/libnfs/_libnfs.so
#7  0x000000000049d18a in call_function.9501 (pp_stack=0x7fffe697d420,
    oparg=<optimized out>) at ../Python/ceval.c:4025
#8  0x000000000048bddc in PyEval_EvalFrameEx (frame=0x216e8f0, throwflag=0)
    at ../Python/ceval.c:2669
#9  0x000000000049d58d in fast_function (nk=<optimized out>,
    na=<optimized out>, n=1, pp_stack=0x7fffe697d5a0, func=0x1f6ded8)
    at ../Python/ceval.c:4111
#10 call_function.9501 (pp_stack=0x7fffe697d5a0, oparg=<optimized out>)
    at ../Python/ceval.c:4046
#11 0x000000000048bddc in PyEval_EvalFrameEx (frame=0x21973d0, throwflag=0)
    at ../Python/ceval.c:2669
```

earlchew:
Perhaps a good invariant to enforce is that each queued callback is invoked exactly once.

sahlberg:
Merged.

PR #217, opened 2017-05-11
Only process pdus once when propagating cancellations or errors

earlchew:


earlchew:
https://github.com/sahlberg/libnfs/issues/215

sahlberg:
Merged, thanks!

PR #220, opened 2017-05-11
Fix crash on timed out mount-nfs calls.

kevin-vigor:
If wait_for_nfs_reply() times out, nfs_mount can return with RPCs
still pending. In that case when the RPCs complete (perhaps because
someone calls destroy_context()), the callbacks run, and private_data
is pointing at what was the stack-allocated cb_data structure. Stack
smashing and segfaulty fun ensue.

Fix by ensuring no RPCs are pending before returning from nfs_mount()
by disconnecting on errors.


sahlberg:
Merged, thanks

PR #224, opened 2017-07-07
Increased resolution for PDU timeout (clock_gettime, only)

doktorstick:
This covers the systems with `clock_gettime`; systems without it still use `time(NULL) * 1000`, which will be subject to the resolution problem this checkin addressed.


sahlberg:
Merged. Thanks.

PR #227, opened 2017-10-02
add cmake build system support

lrusak:
I've added cmake build system support. I hope you find this useful. I took this on to help some of the other Team Kodi windows guys as windows does not support autotools but does support cmake.

I've only tested this on linux (needs testing on windows, osx, solaris, and whatever aros is)

It contains all the functionality (plus more) that autotools did.

You can now use the following build options
```
BUILD_SHARED_LIBS
ENABLE_DOCUMENTATION
ENABLE_EXAMPLES
ENABLE_TESTS
ENABLE_UTILS
```

I did not add support for the `compile_rpc` part as I don't quite understand what it is for and when it is called but it seems fine without it. It should be trivial to add when I know what it is for though.

Also, this installs the following libraries
```
libnfs.so -> libnfs.so.11.0.0
libnfs.so.11.0.0 -> libnfs.so.2.0.0
libnfs.so.2.0.0
```
compared to the old autotools
```
libnfs.so -> libnfs.so.11.0.0
libnfs.so.11 -> libnfs.so.11.0.0
libnfs.so.11.0.0
```
Let me know if this is ok otherwise I will change it.

You can build as follows

```
mkdir -p build
cd build
cmake -DBUILD_SHARED_LIBS=ON
      -DCMAKE_INSTALL_PREFIX=/usr/local
      -DENABLE_EXAMPLES=ON
      -DENABLE_UTILS=ON
      -DENABLE_TESTS=ON
      -DENABLE_DOCUMENTATION=ON
      ..
make -j9
make test
```

Please let me know if you have any questions or concerns

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ stamp-h1
 
 libnfs.pc
 !libnfs.pc.in
+build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,97 @@
+cmake_minimum_required(VERSION 3.2)
+
+project(libnfs
+        LANGUAGES C
+        VERSION 2.0.0)
+
+set(SOVERSION 11.0.0 CACHE STRING "" FORCE)
+
+set(INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib" CACHE PATH "Installation directory for libraries")
+set(INSTALL_INC_DIR "${CMAKE_INSTALL_PREFIX}/include" CACHE PATH "Installation directory for headers")
+set(INSTALL_MAN_DIR "${CMAKE_INSTALL_PREFIX}/share/man" CACHE PATH "Installation directory for manual pages")
+set(INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_PREFIX}/lib/pkgconfig" CACHE PATH "Installation directory for pkgconfig (.pc) files")
+set(INSTALL_CMAKE_DIR "${CMAKE_INSTALL_PREFIX}/lib/cmake/libnfs" CACHE PATH "Installation directory for cmake (.cmake) files")
+
+option(BUILD_SHARED_LIBS "Build shared libraries" ON)
+option(ENABLE_TESTS "Build and run test programs" OFF)
+option(ENABLE_DOCUMENTATION "Build Documentation" OFF)
+option(ENABLE_UTILS "Build util programs" OFF)
+option(ENABLE_EXAMPLES "Build example programs" OFF)
+
+if(ENABLE_TESTS)
+  set(ENABLE_UTILS ON CACHE BOOL "Building utils required by tests" FORCE)
+endif()
+
+include(cmake/Macros.cmake)
+include(cmake/ConfigureChecks.cmake)
+
+include_directories(${CMAKE_CURRENT_BINARY_DIR}
+                    include
+                    include/nfsc)
+
+set(CORE_LIBRARIES nfs)
+set(core_DEPENDS "" CACHE STRING "" FORCE)
+
+if(CMAKE_SYSTEM_NAME STREQUAL Windows)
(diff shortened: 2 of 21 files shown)


sahlberg:
Looks good to me. Merged.

As for windows, we already support for visual studio project files for libnfs.
This was originally contributed by the kodi windows folks.

Makes sense though in that you can now use the same build system for both windows and everything else.

lrusak:
Awesome!

I'm not 100% sure how windows stuff with kodi works, but IIRC they are just using a cmake patch to build anyway, https://github.com/xbmc/kodi-deps/blob/externalProject/patches/libnfs.diff

then they would build it and upload the precompiled binary to their servers where it would just be dll'd at runtime.

A couple more questions,
 - Are you interested in dropping autotools support when this is working on all platforms?
 - Do you want me to do the same thing for libsmb2?

sahlberg:
I am not really willing to drop autotools support but I am happy to
maintain cmake in addition to it, so the users can chose what buildsystem
they want to use.

Libnfs is fairly stable right now as any new development is focused on the
NFSv4 support, so I do not expect there to be much
maintenance overhead to maintain another build system.


If it would be useful for kodi to have cmake support for libsmb2, please do
so.

lrusak:
@sahlberg can you explain to me about the `compile_rpc` function that is present in autotools?

sahlberg:
Ah, the compile_rpc target is just for my personal use.

Normally in ONC-RPC, the protocol compiler (rpcgen) and the
runtime (libxdr and librpc) go hand in hand. They are written in tandem so
that a specific rpcgen
implementation will only work with the specific librpc implementation it is
paired with.

This means, that generally you can not run rpcgen and compile a .x file
into .c and .h on , say, linux, then copy these files
and compile and link the .c and .h files against the librpc that is shipped
on BSD :-(

Thus in order to build from .x files, you generally need to compile the .x
file using rpcgen on teh same platform as where you will compile and link
with the system librpc :-(



Originally libnfs only implemented the upper layer of RPC and then used the
system provided
librpc for the lower layer. That was painful because some platforms do not
have a librpc port, such as Android,
iOS or Windows.

At some stage, to help the kodi folks get libnfs support on Windows, I did
manage to get an old and abandoned windows port of rpcgen and librpc
to build and run on Windows, but it was painful.
In really old versions of the kodi sources you can probably still see
references to "rpcgen.exe" and "librpc.dll". God that was painful.

More painful was that even on linux/bsd/solaris there were minor
differences in the API with subtle differences
making it really painful to support :-(


At some stage I decided that the least painful path forward was to write my
own libxdr and librpc replacement
and stop depending on the paltform provided rpcgen/librpc.
As a benefit that would also make libnfs suddenly work on platforms where
no rpcgen/librpc is even available. Such as I think Android and iOS.

This is what is called ZDR in libnfs. It is my own version of a small
subset of XDR, just enough of it to get the features that libnfs needs.


However, I did not want to write a compiler too, and thus decided to make
libnfs ship "pre-compiled" protocol .h and .c files.
While libnfs does also ship the original protocol definitions in the .x
files,   the .x files are not used during the build
and thus allows people to build even without having a rpcgen protocol
compiler installed.


My ZDR and rpc runtime is slightly different from the one that ships for
example as part of linux glibc so when (very rarely) I need to update the
protocol definitions
and re-run the compiler, I need to do some small "tweaking" of the
generated files so that they will work with libnfs and also to make sure
all compiler warnings are
removed, so that I can use -Wall and -Werror.   This is what the
"compile_rpc" targets are for.

It is just so that I can use the ancient rpcgen compiler that my linux
distro uses, the tweak the output slightly.
Fun fact: rpcgen is so ancient that it still generates K&R style C code :-)

I.e.   "compile_rpc" is only used, manually, when any of the .x files
change, which is very very rare.


Makes sense?

regards
ronnie sahlberg

lrusak:
makes perfect sense, thanks.

I'll leave it for now but if you ever decide to drop autotools just ping me and I'll do it in cmake :+1:

PR #236, opened 2018-01-13
Define F_RDLCK for windows builds

Daniel-Abrecht:
The current libnfs master branch didn't build anymore with mingw+msys since commit "50068b84329fe6e9b8c972fcc22c0c1886775b4b NFSv4: add support for fcntl locking", because F_RDLCK isn't defined in that build environment. This commit defines F_RDLCK if it's not already defined.


sahlberg:
Merged. Thanks.

PR #240, opened 2018-02-03
status variable for the callback function is error

cosmoer:
When a function fails, its status for the callback function should be passed RPC_STATUS_ERROR,because the original status variable is RPC_STATUS_SUCCESS.

Signed-off-by: cosmoer <cosmoer@qq.com>


sahlberg:
Merged, thanks!

PR #246, opened 2018-04-23
Avoid underflow when offset < NFS_BLKSIZE

sirshreyas:


sahlberg:
Hi,

I think I have merged all your patches now.
Many thanks for making libnfs better.


I am always curious about what use-cases the libnfs consumers use and your
comment that you dont use nfs_open() sounds intriguing.


Please, if you want to, tell me your usecase and how you use libnfs :
ronniesahlberg@gmail.com


regards
ronnie sahlberg

PR #249, opened 2018-04-23
nfs_opendir_cb should not queue a READDIR on error

shreyassiravara:
Explanation of the bug:
- nfs_opendir_cb() queues a READDIR when it receives RPC_STATUS_ERROR.
- rpc_purge_all_pdus() explicitly says that no further pdus should be
  queued when rpc_purge_all_pdus() is invoked and the outqueue is being
  cleared.
- Since nfs_opendir_cb() is called in rpc_purge_all_pdus() with
  status=RPC_STATUS_ERROR, this invariant is broken.

Fix:
- Invoke check_nfs3_error() which will invoke the appropriate
  callback with the right error.
- Disallow queueing a request in the if block


sahlberg:
Merged, thanks.

PR #250, opened 2018-04-23
Allow nfs_closedir to accept a nullptr for the nfs context

shreyassiravara:
In the case where dircache is not enabled, we should be able to call
closedir without an nfs context.


sahlberg:
I am travelling this week but will look into this when I return.

sahlberg:
It doesn't hurt to allow calling this function with nfs==NULL
but is there really a valid usecase to call with nfs==NULL ?

Merged anyway. Thanks.

sahlberg:
Merged, thanks.

PR #252, opened 2018-04-27
Add & export a call to initialize the pagecache

shreyassiravara:
For filehandles that are managed outside of the libnfs "high-level" API (i.e., if we don't use nfs3_open()), sometimes we want to initialize the pagecache manually.


sahlberg:
Merged, thanks.

On Fri, Apr 27, 2018 at 11:16 AM, Shreyas Siravara <notifications@github.com

PR #260, opened 2018-07-02
Wrap use of SOCK_CLOEXEC with __linux__

pingerino:
This guards against the case where your clibrary defines SOCK_CLOEXEC
but the underlying network library does not support it.

I'm not sure this is the ideal solution, for context: I'm trying to use libnfs in an embedded context with musl libc and picotcp. Musl libc defines SOCK_CLOEXEC, but picotcp does not support the option.

### DIFF
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -93,12 +93,14 @@ static int
 create_socket(int domain, int type, int protocol)
 {
 #ifdef SOCK_CLOEXEC
-	/* Linux-specific extension (since 2.6.27): set the
+#ifdef __linux__
+    /* Linux-specific extension (since 2.6.27): set the
 	   close-on-exec flag on all sockets to avoid leaking file
 	   descriptors to child processes */
 	int fd = socket(domain, type|SOCK_CLOEXEC, protocol);
 	if (fd >= 0 || errno != EINVAL)
 		return fd;
+#endif
 #endif
 
 	return socket(domain, type, protocol);


sahlberg:
Might be better with a single #if defined...   defined
than two nested ifdefs, but this is good enough.


Thanks for your contribution.  Is it worth adding something about your platform to README?
Second, please have a look at libsmb2 too. If you want to use nfs from userspace, chances are good you might want to use smb2 too :-)

pingerino:
@sahlberg thanks, I'll check out libsmb2. The current work is in the context of our advanced operating systems course, so it's not technically a supported platform for libnfs. However if we do add more concrete support for libnfs to the seL4 user-level infrastructure, we'll consider updating the README. Prior to this year, we had been using an old nfs library which only supported up to NFSv2. 

@Daniel-Abrecht to answer your questions, we don't have an OS. The port is currently for a student project to build an OS, which builds a basic microkernel based operating system on top of seL4. We provide picotcp and libnfs as building blocks for their network mounted file system and TCP/UDP traffic. muslc may not be the best choice, but it is the only c library we currently have ported to seL4, and our team doesn't have the bandwidth to consider other ports at this point. I'm more than happy for this change to be backed out if is isn't the right solution, and I'll find another work around. 

If anyone's curious, the project is here: https://www.cse.unsw.edu.au/~cs9242/17/project/index.shtml, although it hasn't been updated for this year yet.

Daniel-Abrecht:
@pingerino interesting, now that I know how everything is used, I think picotcp  and muslc have been a really good choice for this use case. And I think this is a really awesome project for students to have.

Since the change may only lead to a few file descriptors getting leaked on some unix systems in some special cases, which usually isn't much of an issue anyway, I think it's reasonable to keep it, even though, thats not my decision to make anyway. But I would still recommend to change the define of SOCK_CLOEXEC and O_CLOEXEC in your muslc port to the constant 0, since there may be other projects using them, and this should work in almost all cases where these constants may be used.

PR #261, opened 2018-07-06
configure.ac: fix --enable-utils

Polynomial-C:
Running "configure --enable-utils" _disables_ compilation/installation of
utils (and vice versa) while omitting the --(en|dis)able-utils option auto-
enables utils.
This patch fixes the logic when the option is given on command line but
keeps utils build enabled when the option was omitted.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -25,11 +25,13 @@ fi
 AC_SYS_LARGEFILE
 
 #option: utils
+MAYBE_UTILS="utils"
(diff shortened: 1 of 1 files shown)


sahlberg:
Hahaha

This is how great my autotools skills are :-)


Thanks.

Polynomial-C:
I find this a very comprehensive page when it comes to basic autotools stuff:

[Autotools Mythbuster](https://autotools.io)

PR #263, opened 2018-07-12
libnfs.c: include <time.h> in libnfs.c

pingerino:
time() is used in this file, which is from time.h. Fixes a warning.


sahlberg:
Merged. Thanks.

PR #267, opened 2018-09-13
fix for dup2 on windows

hdornemann:
dup2 doesn't work with sockets on windows.  It seems like we can just close out the old socket and start using the newly connected one instead.


sahlberg:
Merged, thanks!

PR #316, opened 2020-05-06
Update libnfs-zdr.c

DetlefGolze:
libnfs_zdr_array() needs to determine element count before calculating size to allocate.


sahlberg:
Merged. Thanks.

PR #318, opened 2020-05-19
Fix NFS4 behavior of chmod/chown/utimes

dmeister:
This patch fixes various issues around chmod/chown/utimes:
- The chmod/chown/utimes approach uses open_async_internal, which only works
  for files. This patch uses lookup_path so that chmod/chown is also
  possible on directories.
- Implements SET_TO_SERVER for utimes

I never succeeded in getting all tests passing even without
these changes. Nor did I succeed in getting the CMake compilation to
work. Sorry. The patch includes extra unit tests for these calls.
We use the libary as part of a test harness to test an NFS server
implementation, so I am reasonably confident that the changes work.
They are also manually tested against the Linux Kernel NFS Server.

### DIFF
--- a/examples/nfs-io.c
+++ b/examples/nfs-io.c
@@ -35,6 +35,7 @@ WSADATA wsaData;
 #include <inttypes.h>
 #include <string.h>
 #include <sys/stat.h>
+#include <sys/time.h>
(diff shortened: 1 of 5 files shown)


sahlberg:
Merged. Thanks!

PR #319, opened 2020-05-19
Allocate larger buffer for write operations in NFS4

dmeister:
The NFS4 pwrite operations contains a memory corruption if a write
is large than 4KB. The reason is that the buffer allocated within
nfs4_compund_async has a size of 4K and it does not account
for the size of the write buffer. To fix this the patch
padds a rpc_allocate_ndu2 function which an additional parameter
as an extra hint for the allocation.


sahlberg:
Merged. thanks!

PR #323, opened 2020-06-04
Implement mount_getexports_timeout

peak3d:
When calling mount_getexport on an unavaible source, especially using windows with default system tcp timeout values lead to very long waiting period.

Because of this Kodi plans to support user defined nfs timeout, which already works fine for all operations using rpc_settimeout(). Timeout in mount_getexports does not work, because wait_for_reply() does not respect the timeout value.

This PR implements a new export function mount_getexports_timeout which is nearly identical to mount_getexports() but let you pass a timeout value as second parameter.


sahlberg:
Merged, thanks.

PR #324, opened 2020-06-22
Fix seekdir

adamyi:
The previous implementation doesn't actually update dir->current but
resets dir->current to dir->entries


sahlberg:
Merged, thanks.

PR #326, opened 2020-07-26
Don't define u_int

asomers:
It's usually defined by the platform in types.h or sys/types.h.
The compiler complains if libnfs redefines it.


asomers:
cc @shivankgarg98

sahlberg:
Merged. Thanks.

PR #372, opened 2022-01-28
fix multithreading mutex

heitbaum:
Without -pthread following error during build:
- nfsv4: fix multithreading mutex


sahlberg:
merged, thanks

PR #374, opened 2022-02-05
Add an opaque pointer argument to service_proc

vitalif:
Without this argument it's impossible to pass an object instance to procedure callbacks which basically forces users to either use global variables or to write additional boilerplate code and find object instances from rpc_context pointer values.

### DIFF
--- a/examples/nfs4-cat-talloc.c
+++ b/examples/nfs4-cat-talloc.c
@@ -680,7 +680,7 @@ void setclientid_cb(struct rpc_context *rpc, int status, void *data,
 /*
  * NULL procedure for the callback protocol.
  */
-static int cb_null_proc(struct rpc_context *rpc, struct rpc_msg *call)
+static int cb_null_proc(struct rpc_context *rpc, struct rpc_msg *call, void *opaque)
 {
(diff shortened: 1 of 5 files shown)


sahlberg:
Merged, thanks!

Can I ask what application you are using the server functionality for? Just curious.

vitalif:
Thanks for merging! :-)
I'm developing an NFS proxy (not a full-featured FS, just an access method for block storage) for Vitastor https://github.com/vitalif/vitastor/ using libnfs, because I found it much easier to use than for example writing a plugin for nfs-ganesha and also because it integrates with an external event loop - and I use my own io_uring based event loop :-) nfs-ganesha APIs aren't 1:1 equivalent to RFC NFS methods, so when writing a plugin you also have to figure out nfs-ganesha's own logic, and with libnfs you can just use RFC as documentation which is very handy. Some nfs-ganesha methods also use threading and I hate multi-threaded code :)) with a single thread I can just ignore most race conditions. I'll probably have to add multiple threads support to my client library later because otherwise it has memory copying bottleneck - but I'll almost definitely use only several threads with eventloops instead of spawning a separate thread per each connection or per each operation...

PR #379, opened 2022-03-15
autotools: Fix building shared library in Windows platform

Biswa96:


Biswa96:
Would it be possible to review this pull request please?

sahlberg:
merged.
if it fails the build I may have to revert it

PR #380, opened 2022-04-03
autotools: fix syntax error in configure when pthreads disabled

thesamesam:
Avoids this warning/syntax error in configure when pthreads aren't
enabled (default):
```
checking for special C compiler options needed for large files... no
checking for _FILE_OFFSET_BITS value needed for large files... no
./configure: 14227: test: =: unexpected operator
checking whether SO_BINDTODEVICE is available... yes
checking whether getpwnam() is available... yes
```

The LHS of the test will be blank, so use an 'x' to give a dummy
value, like we do elsewhere.

Signed-off-by: Sam James <sam@gentoo.org>


sahlberg:
Thanks. I am very much not good at autotools  or configure so again   THANKYOU for fixing my mistakes!

PR #382, opened 2022-06-18
libnfs.h: remove comment about non-functional nfsv4 support

lrusak:
Just something I noticed. I saw that a similar comment was dropped in d86d5d18514a25da0347f5bc84aabac65302852e


sahlberg:
Merged. Thanks

PR #390, opened 2022-08-09
Cleanup

FGasper:
Compilation on macOS yields quite a number of warnings. This silences those.

### DIFF
--- a/utils/nfs-stat.c
+++ b/utils/nfs-stat.c
@@ -47,6 +47,7 @@ WSADATA wsaData;
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdint.h>
+#include <inttypes.h>
 #include <sys/types.h>
 #include <fcntl.h>
 #include "libnfs.h"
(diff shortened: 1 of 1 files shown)


sahlberg:
libnfs-raw-nfs.c is autogenerated by rpcgen  so we can't/shouldn't edit it directly.

FGasper:
> libnfs-raw-nfs.c is autogenerated by rpcgen so we can't/shouldn't edit it directly.

Ahh ok. I saw the copyright at the top and thought otherwise. Re-pushed without that change.

sahlberg:
Merged. thanks!

Nable80:
https://github.com/sahlberg/libnfs/pull/390/commits/94433c80339c9156c80d6411fc418900bed59759#diff-3798c2f1ec1153258b8fe223a0c917fced30ef20baaa6b43fce1e49f24cb6ae1L286-R289

I think there's a reason to show warnings for these lines. At least it'll break on big-endian machines with 32-bit ``time_t`` (old Macs?), although they are almost extinct now. Other changes may be tricky too.

Is it possible to switch to ``time_t`` (or even ``struct timespec``) in ``struct nfs_stat_64`` or it'll bring much more problems?

PR #398, opened 2022-10-11
Expose further configuration options

rovo89:
The options auto-traverse-mounts, nfsport and mountport could be set via URL, but not from libaries.

### DIFF
--- a/include/nfsc/libnfs.h
+++ b/include/nfsc/libnfs.h
@@ -299,8 +299,11 @@ EXTERN void nfs_set_pagecache(struct nfs_context *nfs, uint32_t v);
 EXTERN void nfs_set_pagecache_ttl(struct nfs_context *nfs, uint32_t v);
 EXTERN void nfs_set_readahead(struct nfs_context *nfs, uint32_t v);
 EXTERN void nfs_set_debug(struct nfs_context *nfs, int level);
+EXTERN void nfs_set_auto_traverse_mounts(struct nfs_context *nfs, int enabled);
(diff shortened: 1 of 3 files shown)


sahlberg:
Merged. thanks.

rovo89:
@sahlberg Thanks for merging this! Please also see my other PR #396 for a significant performance improvement.

One more question: I thought `auto-traverse-mounts` would skip "sub-mounts". On NetApp filers there's an option to have a `.snapshot` subdirectory where you can browse previous versions. Information from `stat` on Linux:

| path | device | inode |
| --- | --- | --- |
| `/` | 76 | 2666725474 |
| `/.snapshot/` | 76 | 2666725443 |
| `/.snapshot/daily.2022-10-10_0200` | 76 | 2666710626 |
| `/.snapshot/daily.2022-10-10_0200/` | 104 | 2666710626 |
| `/.snapshot/daily.2022-10-10_0200` | 104 | 2666710626 |
| `/.snapshot/daily.2022-10-10_0200/README` | 104 | 2666745996 |

After the `stat` on `/.snapshot/daily.2022-10-10_0200/` (with trailing slash) a new entry appears in the `mount` command with that server path. So it looks like a different mount which is automatically mounted.

However `nfs-ls -R nfs://xxx/yyy/zzz?auto-traverse-mounts=0` also lists the contents of all `.snapshot` subdirectories. Is this a bug? Is there any other option to avoid this? If not, would it be worth opening a feature request for this? Or is it in the application's responsibility to prune recursion for these cases?

PR #400, opened 2022-10-13
Reduced memory allocations.

davidgaleano:
Removed one memory allocation per PDU creation. The output data memory block is now behind the header block.

Removed one memory allocation per response. The input buffer is now kept and reused, automatically growing to whatever size is required.


sahlberg:
This looks fine.  Thanks.

PR #401, opened 2022-11-01
multithreading.c: wait until poll signals an event

pborzenkov:
0 as timeout makes poll() return as soon as possible, even if there are no events to process. This, in turn, results in the service thread consuming the whole CPU doing nothing.

Fix by allowing poll() to sleep until there is an event to signal.

### DIFF
--- a/lib/multithreading.c
+++ b/lib/multithreading.c
@@ -81,7 +81,8 @@ static void *nfs_mt_service_thread(void *arg)
 		pfd.events = nfs_which_events(nfs);
 		pfd.revents = 0;
 
-		ret = poll(&pfd, 1, 0);
+		// Wake up at least every 100ms to process timeouts.
+		ret = poll(&pfd, 1, 100);
 		if (ret < 0) {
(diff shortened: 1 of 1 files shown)


sahlberg (changes requested):
We can not wait indefinitely in the poll() since that would break timeout processing.
We thus need to make sure we call the service routine at least on a semi-regular basis.

These timeouts do not need to be super accurate so we could get away with a rather large timeout value for poll().
Can you change the patch to have poll() only block for, say 1 second, and test?
Only running this loop once per second should drop the cpu processing tiime as well as still allow timeout processing to occur.

pborzenkov:
> We can not wait indefinitely in the poll() since that would break timeout processing. We thus need to make sure we call the service routine at least on a semi-regular basis.

@sahlberg Thanks for your review! I totally forgot about the timeouts. I will set the `poll` timeout to 100ms. This is small enough to process all the timeouts in a timely manner, but also shouldn't put any strain on the CPU.

sahlberg (approved):

PR #403, opened 2022-11-19
Fix deprecation warnings on modern macOS

pborzenkov:
syscall() and POSIX semaphores are deprecated on modern macOS. Thus, libnfs can't be compiled with -Werror and multi-threading there.

To fix this, replace POSIX semaphores with the ones from libdispatch (available since 10.6 as part of the standard SDK) and syscall(gettid) with non-portable pthread function.

Both changes are behind autoconf check, so hopefully won't break anything.


pborzenkov:
Hmm, both Cygwin and OpenBSD failures seem to be unrelated to the changes made by this PR.

sahlberg:
Merged, thanks

PR #404, opened 2022-11-24
Add options to set dircount/maxcount for readdir()

rovo89:
I did more tests and think that both dircount and maxcount should be configurable separately (although there is a shortcut to set both to the same value). Some servers seem to ignore dircount completely, on others the ratio is more like 1:3 than 1:8, so this enables fine-tuning depending on the actual scenario.


rovo89:
The check failures seem to be caused by infrastructure issues, not by my changes...

sahlberg:
Merged, thanks

PR #408, opened 2022-12-01
Fix integer size for readdir's dircount/maxcount

rovo89:
I had misinterpreted the types, the NFS protocol has only 4 bytes for these fields.


sahlberg:
Merged, thanks

PR #409, opened 2022-12-01
Use uint32_t for uid/gid in authentication and chown

rovo89:
They're defined as unsigned int in RFC 5531 and RFC 1813. Some parts of the code (e.g. libnfs_authunix_create()) already used uint32_t, causing implicit casts.

I didn't touch rquota since I couldn't find an RFC for it and rquota.h seems to use int.

### DIFF
--- a/examples/nfs-io.c
+++ b/examples/nfs-io.c
@@ -154,8 +154,8 @@ int main(int argc, char *argv[])
 			fprintf(stderr, "Invalid arguments for chown");
 			goto finished;
 		}
-		int uid = strtol(argv[2], NULL, 10);
-		int gid = strtol(argv[3], NULL, 10);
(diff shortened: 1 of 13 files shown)


rovo89:
I'm not 100% sure about compatibility for this one, I hope you can better estimate the impact.

Implicit casts should continue to work, at least it was possible to call `nfs_set_uid(nfs, 4000000000)` (i.e. a value which is bigger than `INT32_MAX` but smaller than `UINT32_MAX`) and it got first converted to `int` (probably as a negative number) and then back to `uint32_t` for the call to `libnfs_authunix_create()`.

sahlberg:
They should be 32 bit already on most platforms that matter.
I don't think this is important enough to change the abi.

PR #414, opened 2022-12-19
Write multiple pdus at once.

dgaleano-ilm:
Attempt to batch the iovectors from several pdus in order to reduce the number of calls to writev.
In the best case this change will reduce the number of calls to writev to an eighth of what they were before.

The maximum number of vectors was reduced to match the limit set on the Linux kernel for the fast code path that handles iovectors.

### DIFF
--- a/include/libnfs-private.h
+++ b/include/libnfs-private.h
@@ -173,7 +173,7 @@ struct rpc_context {
         struct rpc_endpoint *endpoints;
 };
 
-#define RPC_MAX_VECTORS 16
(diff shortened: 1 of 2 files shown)


dgaleano-ilm:
The existing code is using a mixture of tabs and spaces so I was not sure what to use but the code I replaced in particular was using more spaces than tabs so I went with spaces. Please let me know if you prefer tabs.

sahlberg:
Merged, thanks.

PR #416, opened 2023-01-05
Patch to support NLM Share

pragyashree-nutanix:
Added the changed required to support NLM4_SHARE


sahlberg:
Merged, thanks.
Ideally we should not do manual changes to libnfs-raw-nlm.c/h since these files are machinegenerated from nlm.x
(cd nlm; make compile_rpc)
but this is an obscure protocol and there is not really any changes/development ever happening to this protocol so
some handwritten additions should be just fine.
Probably not worth the effort to make sure they can be machine-generated.

PR #423, opened 2023-06-15
Check return value from adjust_inbuf.

dgaleano-ilm:
Otherwise we may have failed to allocate the buffer required to read the rest of the response and we would not know, resulting in reading the incoming data into a buffer too small for it.


sahlberg:
Merged, thanks

PR #425, opened 2023-06-20
Handle small reads for record marker field.

dgaleano-ilm:
If the call to recv for the record marker reads less than 4 bytes then the state will still be READ_RM and the next call to recv will overwrite the lowest bytes of the variable "record_marker", which when converted to little-endian will generate an incorrect and very high value for the payload.

The code is changed to always use the "inpos" value to move the buffer forward to the correct location for the next bytes, previously it was only used for the payload buffer. Note that the size of the data to read from the variable "pdu_size" was already usign "inpos".


sahlberg:
A, good catch. thanks!

PR #427, opened 2023-07-05
Always close the socket.

dgaleano-ilm:
If the connection process fails it is possible for "is_connected" to be zero whilst still having a valid socket handle on "fd". The previous code was keeping the socket open on those situations and any future calls to "rpc_connect_async" will all fail with "Trying to connect while already connected", the only workaround was to destroy the context and start again. By always closing the socket if we have one we avoid that issue.


sahlberg:
Nice catch. thanks!

PR #428, opened 2023-07-07
Allow setting a limit on the number of commands in-flight.

dgaleano-ilm:
Some NFS servers will limit the number of in-flight requests from NFS clients to a maximum of 128.
This limit is configured as "sunrpc.tcp_max_slot_table_entries". When a client reaches that limit the server responds by decreasing the window size for the TCP connection, hence decreasing throughput until the window size recovers.

The new functionality will allow applications to set a limit on the number of in-flight requests (or "awaiting" as the API defines them).

Also, a new function was added to request the number of commands waiting because this number may be very different to the total size of the output queue and it could be useful to know the difference between the two.

### DIFF
--- a/include/libnfs-private.h
+++ b/include/libnfs-private.h
@@ -139,6 +139,8 @@ struct rpc_context {
         uint32_t num_hashes;
 	struct rpc_queue *waitpdu;
 	uint32_t waitpdu_len;
+	uint32_t max_waitpdu_len;
+
 #ifdef HAVE_MULTITHREADING
         int multithreading_enabled;
         libnfs_mutex_t rpc_mutex;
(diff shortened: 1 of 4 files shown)


sahlberg:
To start, I have no major objections to something like but as there is no flow-control in ONC-RPC or NFS,  having to add this sounds like more of a major server bug than anyting else. A server without flow control should just be able to 
"stop reading more request PDUs from the socket" server-side to deal with that.

Out of curiosity, what server implementation has trouble with this? You can mail me privately to not shame them.

You use 0xffffffff as the no-limit.
Can you change the patch to use 0 for max as "no limit" to match other code that does "max or not limit".
Then I will merge it.

dgaleano-ilm:
I will change the code to use zero as the no-limit.

dgaleano-ilm:
I have changed the code to use zero as no-limit.

sahlberg:
Merged, thanks

PR #429, opened 2023-08-12
ci: Run tests on Linux

rbalint:
This PR enables building examples, tests and running tests.

Building examples and many tests fail though, thus it would probably make sense to fix those issues first.

Having examples that build and having tests that pass would be a nice for distribution maintainers and I guess for other developers, too.


sahlberg:
The workflow had a failure. Can you look into it?

rbalint:
I've fixed building the examples, but some of the NFSv4 tests still fail (for a long time, I believe).
Please look into those failures.

PR #451, opened 2024-03-28
PS3: initialize addrinfo struct

cy33hc:
Apply the same fix for libsmb2. It hasn't caused an issue yet, but just isolate any issue down the road.


sahlberg:
Merged. thanks!

Issue #455, opened 2024-04-11
Build error on macOS: `error: 'IFNAMSIZ' undeclared here (not in a function)` etc.


barracuda156:
With modern GCC there are _a lot_ of these `buf`-related warnings, which eventually turn into error:
```
/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_libnfs/libnfs/work/libnfs-5.0.3/nfs/libnfs-raw-nfs.c: In function 'zdr_MKDIR2resok':
make[2]: *** [mount/CMakeFiles/nfs_mount.dir/mount.c.o] Error 1/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_libnfs/libnfs/work/libnfs-5.0.3/nfs/libnfs-raw-nfs.c:2877:27: warning: variable 'buf' set but not used [-Wunused-but-set-variable]
 2877 |         register int32_t *buf;
      |                           ^~~

. . .
In file included from /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_libnfs/libnfs/work/libnfs-5.0.3/portmap/portmap.c:25:
/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_libnfs/libnfs/work/libnfs-5.0.3/include/libnfs-private.h:178:21: error: 'IFNAMSIZ' undeclared here (not in a function)
  178 |         char ifname[IFNAMSIZ];
      |                     ^~~~~~~~
. . .
```
Old gcc-4.2 is more forgiving, but it fails on the second issue:
```
In file included from /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_libnfs/libnfs/work/libnfs-5.0.3/nsm/nsm.c:27:
/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_libnfs/libnfs/work/libnfs-5.0.3/include/libnfs-private.h:178: error: ‘IFNAMSIZ’ undeclared here (not in a function)
[...]

PR #456, opened 2024-04-12
Fix for older macOS: include sys/socket.h before net/if.h

barracuda156:
Closes: https://github.com/sahlberg/libnfs/issues/455

@sahlberg Could you please take a look, if this works for you?


sahlberg:
Merged, thanks!

PR #457, opened 2024-04-25
Add RPC-with-TLS support

linuxsmiths:
This PR adds RPC-with-TLS (RFC 9289) support to libnfs. It depends on the following:
- GnuTLS version 3.4.6+
- Linux kernel 5.10+

It adds a new mount option - xprtsec=[none,tls,mtls] (similar to Linux NFS client)

I've currently enabled it only on Linux since that's what my target system is. BSD should be easy to make work if someone has a system they can try out.
Also, this change only adds client support.
I've only made changes to the CMake build files.

Tested to make sure it works fine with reconnects.

Sample command:

./examples/nfs-writefile /mnt/random 'nfs://nfsserver/mnt/testfile?version=3&xprtsec=tls&debug=1'

Performance testing:
With AES-128-GCM there's almost no overhead and I could get ~1GBps/core which is same as w/o TLS.
This is because kTLS does the encryption while copying data from user to kernel (so there's no extra copy) and encryption is accelerated by the CPU.

### DIFF
--- a/lib/libnfs.c
+++ b/lib/libnfs.c
@@ -169,6 +169,19 @@ nfs_set_security(struct nfs_context *nfs, enum rpc_sec sec)
 #endif
 }
 
+#ifdef HAVE_TLS
+void
+nfs_set_xprtsecurity(struct nfs_context *nfs, enum rpc_xprtsec xprtsec)
+{
+	/* Ensure only permissible values are being set */
+	assert(xprtsec == RPC_XPRTSEC_NONE ||
+	       xprtsec == RPC_XPRTSEC_TLS ||
+	       xprtsec == RPC_XPRTSEC_MTLS);
+
+	nfs->rpc->wanted_xprtsec = xprtsec;
+}
+#endif /* HAVE_TLS */
+
 int
 nfs_get_fd(struct nfs_context *nfs)
 {
@@ -207,7 +220,7 @@ nfs_get_error(struct nfs_context *nfs)
                         }
                 }
         }
-#endif        
+#endif
 	return nfs->error_string ? nfs->error_string : "";
 };
 
@@ -259,6 +272,19 @@ nfs_set_context_args(struct nfs_context *nfs, const char *arg, const char *val)
 		} else {
 			nfs_set_readdir_max_buffer_size(nfs, atoi(val), atoi(val));
 		}
+#ifdef HAVE_TLS
+	} else if (nfs->rpc && !strcmp(arg, "xprtsec")) {
+		if (!strcmp(val, "none")) {
+			nfs_set_xprtsecurity(nfs, RPC_XPRTSEC_NONE);
+		} else if (!strcmp(val, "tls")) {
+			nfs_set_xprtsecurity(nfs, RPC_XPRTSEC_TLS);
+		} else  if (!strcmp(val, "mtls")) {
+			nfs_set_xprtsecurity(nfs, RPC_XPRTSEC_MTLS);
+		} else {
+			nfs_set_error(nfs, "Unknown/unsupported xprtsec type : %s", val);
+			return -1;
+		}
+#endif /* HAVE_TLS */
 #ifdef HAVE_LIBKRB5
 	} else if (nfs->rpc && !strcmp(arg, "sec")) {
                 /*
@@ -346,7 +372,7 @@ nfs_parse_url(struct nfs_context *nfs, const char *url, int dir, int incomplete)
                 strcpy(strp + 1, strp + 3);
                 strp++;
         }
-       
+
 	if (urls->server[0] == '/' || urls->server[0] == '\0' ||
 		urls->server[0] == '?') {
 		if (incomplete) {
@@ -433,7 +459,8 @@ flags:
 		if (strp2) {
 			*strp2 = 0;
 			strp2++;
-			nfs_set_context_args(nfs, strp, strp2);
+			if (nfs_set_context_args(nfs, strp, strp2) != 0)
+				return NULL;
 		}
 	}
 
@@ -447,6 +474,21 @@ flags:
 		free(urls->server);
 		urls->server = NULL;
 	}
+
+#ifdef HAVE_TLS
+	/*
+	 * Call this in the end after all options are processed, as it uses
+	 * rpc->debug.
+	 */
+	if (nfs->rpc->wanted_xprtsec != RPC_XPRTSEC_NONE) {
+		/* tls_global_init() MUST succeed for us to use TLS security */
+		if (tls_global_init(nfs->rpc) != 0) {
+			nfs_set_error(nfs, "tls_global_init() failed!");
+			nfs_destroy_url(urls);
+			return NULL;
+		}
+	}
+#endif
 
 	return urls;
 }
@@ -498,7 +540,7 @@ nfs_init_context(void)
 		return NULL;
 	}
 	memset(nfsi, 0, sizeof(struct nfs_context_internal));
-        
+
 	nfs = malloc(sizeof(struct nfs_context));
 	if (nfs == NULL) {
                 free(nfsi);
@@ -544,6 +586,7 @@ nfs_init_context(void)
         nfs_mt_mutex_init(&nfs->nfsi->nfs4_open_counter_mutex);
         nfs_mt_mutex_init(&nfs->nfsi->nfs4_open_call_mutex);
 #endif /* HAVE_MULTITHREADING */
+
 	return nfs;
 }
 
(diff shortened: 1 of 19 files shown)


sahlberg:
Very nice. I will review this over the weekend.
Until then, can you add a licence boilerplate to the two new files?
LGPLv2.1 or compatible.  I am fine with BSD licence too.

lib/libnfs.c
```diff
@@ -2238,3 +2383,59 @@ rpc_null_task(struct rpc_context *rpc, int program, int version, rpc_cb cb,
...
+
+#ifdef HAVE_TLS
+/*
+ * Call this in place of rpc_null_task() if user wants TLS security for the RPC
+ * connection. Since we only ever send AUTH_TLS NULL RPC for NFS_PROGRAM it does
+ * not need the program parameter.
+ * Callback 'cb' is called not after we get the reply for this AUTH_TLS NULL RPC,
+ * but after the TLS handshake completes (success or failure).
+ */
+struct rpc_pdu *
```
sahlberg, line 2395: ONC-RPC over TLS can in theory be used also for other protocols than nfs 4.x, yes?

In that case, would it make sense to add a 'int program' argument here
just so that if someone in the future that wants to add TLS support for other protocols?
Or is that NFSv4.x is unique in the way that it does not rely on portmapper the reason?
I ask from ignorance since I don't know how/if running PORTMAPPER would even be possible across TLS.

Not asking for a change, only to educate myself.

lib/libnfs.c
```diff
@@ -2238,3 +2383,59 @@ rpc_null_task(struct rpc_context *rpc, int program, int version, rpc_cb cb,
...
+	}
+	data->cb 	   = cb;
+	data->private_data = private_data;
+
+	/*
+	 * Set MSbit in procedure number to convey use of AUTH_TLS
+	 * This should not interfere with valid procedure numbers as they are all
+	 * small numbers.
+	 */
+	pdu = rpc_allocate_pdu(rpc, NFS_PROGRAM, nfs_version, (0 | (1U << 31)),
```
sahlberg, line 2420: In rpc_allocate_pdu2() we use 0x80000000U to describe this.  Can you change (1 <<31)  to that so we have the same constant/expression in both places?

lib/nfs_v3.c
```diff
@@ -1128,6 +1129,11 @@ nfs3_mount_async(struct nfs_context *nfs, const char *server,
         free(nfs->nfsi->server);
 	nfs->nfsi->server = new_server;
         
+#ifdef HAVE_TLS
+	free(nfs->rpc->server);
+	nfs->rpc->server = strdup(nfs->nfsi->server);
```
sahlberg, line 1134: Check for NULL?

lib/nfs_v4.c
```diff
@@ -1625,6 +1625,11 @@ nfs4_mount_async(struct nfs_context *nfs, const char *server,
         free(nfs->nfsi->server);
         nfs->nfsi->server = new_server;
 
+#ifdef HAVE_TLS
+	free(nfs->rpc->server);
+	nfs->rpc->server = strdup(nfs->nfsi->server);
+#endif
```
sahlberg, line 1631: check for NULL?

lib/pdu.c
```diff
@@ -465,10 +500,36 @@ int rpc_queue_pdu(struct rpc_context *rpc, struct rpc_pdu *pdu)
...
+	 * Another case where we send straight away is the AUTH_TLS NULL RPC.
+	 * This is particularly important for the reconnect case where we want to
+	 * ensure TLS handshake completes successfully before we can send any of
+	 * the queued RPCs waiting. If we do not send here this AUTH_TLS NULL
+	 * RPC will need to be queued before all other waiting RPCs and even then
+	 * we need to be careful that we don't send any of those RPCs till the
+	 * TLS handshake is completed and the connection is secure.
+	 * Sending inline here makes the handling simpler in rpc_service().
+	 */
+	if (rpc->is_udp != 0
```
sahlberg, line 515: Personal style,
I would find
#ifdef HAVE_TLS
if (rpc->is_udp != 0 || pdu->expect_starttls) {
#else
if (rpc->is_udp != 0) {
#endif
easier to read.

Don't change it, it is fine. Just a personal comment.

sahlberg (changes requested):
Looks absolutely great.  Three minor nits, if you fix them I will check it in.

sahlberg (changes requested):
Please have a look at the build failures too. OSX and Windows look like they might just be unrelated environmental issues since they shouldn't be using TLS anyway (and the compile would fail if they did).

The linux failure though,  does it need an update to the .github/workflows/....  file to force disabling TLS support?   
As TLS will not work without certs being set up, should the build default to build without TLS and only build with the support if --enable-tls  is used?

lib/libnfs.c
linuxsmiths: Technically since it secures the RPC transport it can be used by any RPC program. Since Linux NFS server is the first implementation of RPC-with-TLS I'm taking the liberty of aligning with its behaviour. Linux NFS server today doesn't use secure RPC channel for anything other than the NFS program, i.e., neither PORTMAP nor MOUNT programs use it, so I'd say let's go with this initially, since at least that makes us interoperable with the only server implementation we can get our hands on. 
Do you take part in any connectathon or any other interop events? Maybe you or someone can then try interop'ing libnfs with other NFS servers that support RPC-with-TLS and based on that input we can make it configurable. For now I'd leave it like this.

lib/nfs_v3.c
linuxsmiths: sorry, which one?
nfs->nfsi->server is guaranteed to not be NULL because of the above code.
since free(NULL) is a no-op, we are fine with nfs->rpc->server being NULL.
nfs is not NULL, nfs->rpc is also being used already so that also cannot be NULL.

lib/pdu.c
linuxsmiths: sure enough, I did have this question whether to do it this way or the way I did.
I chose the way I did as it's only slightly "less readable" if we may call it but it is less prone to both tls and non-tls code evolving differently. I'll leave this. Thanks.

linuxsmiths:
> Please have a look at the build failures too. OSX and Windows look like they might just be unrelated environmental issues since they shouldn't be using TLS anyway (and the compile would fail if they did).
> 
> Please check the linux failure though.



> Please have a look at the build failures too. OSX and Windows look like they might just be unrelated environmental issues since they shouldn't be using TLS anyway (and the compile would fail if they did).
> 
> Please check the linux failure though.

I don't see this failure when I try locally.. could it be transient? Have you seen it fail before?

/home/linuxsmiths_libnfs/tls-change1/libnfs# sudo bash -c "cd build/tests;./test_0100_ls_basic.sh"
basic ls test
Testing nfs-ls on root of export ... [OK]
Create a file and verify nfs-ls can see it ... [OK]

lib/nfs_v3.c
sahlberg: Ok, I did not check for that, if it is fine if nfs->rpc->server might be NULL later if strdup fails. Then that is fine.
Thanks for explaining.

sahlberg:
> > Please have a look at the build failures too. OSX and Windows look like they might just be unrelated environmental issues since they shouldn't be using TLS anyway (and the compile would fail if they did).
> > Please check the linux failure though.
> 
> > Please have a look at the build failures too. OSX and Windows look like they might just be unrelated environmental issues since they shouldn't be using TLS anyway (and the compile would fail if they did).
> > Please check the linux failure though.
> 
> I don't see this failure when I try locally.. could it be transient? Have you seen it fail before?
> 
> /home/linuxsmiths_libnfs/tls-change1/libnfs# sudo bash -c "cd build/tests;./test_0100_ls_basic.sh" basic ls test Testing nfs-ls on root of export ... [OK] Create a file and verify nfs-ls can see it ... [OK]

I have not seen this failure before but I accept it might be environmental or transient.
If it returns I will debug it and fix it.

sahlberg:
Clicked the wrong button. I meant to resolve the discussion, not close the pull request. Sorry.

It is merged now. Thank you a lot for this contribution. This is great.

sahlberg:
On Fri, 26 Apr 2024 at 18:40, linuxsmiths ***@***.***> wrote:

> Please have a look at the build failures too. OSX and Windows look like
> they might just be unrelated environmental issues since they shouldn't be
> using TLS anyway (and the compile would fail if they did).
>
> Please check the linux failure though.
>
> Btw, did you notice that I haven't update the autobuild files as I'm not
> familiar with that.
> Do we plan to support both or just the cmake?
>

I didn't actually go back and check.
Don't worry about it. I am much more familiar with autotools than CMake so
I can easily add autotool support to it later.
No problem.
Thankyou a lot for this contribution!

PR #458, opened 2024-04-29
Added (more) TCP and RPC resiliency.

linuxsmiths:
This change adds the following resiliency to libnfs clients.

1. Set TCP keepalive to recognize servers that silently go away or stop responding to TCP.
2. Add RPC timeout and retransmit support. Added following two mount options to control RPC retransmit behaviour:
   - timeo=< int >
   - retrans=< int >
   As we can see these are deliberately named similar to the Linux NFS mount options for ease of use/understanding.

Testing done:
Tested by blocking TCP usign iptables to induce retransmit and make sure data xfer is valid.
Various stress runs while constantly killing/blocking connections.

### DIFF
--- a/include/nfsc/libnfs.h
+++ b/include/nfsc/libnfs.h
@@ -336,6 +336,7 @@ EXTERN void nfs_set_debug(struct nfs_context *nfs, int level);
 EXTERN void nfs_set_auto_traverse_mounts(struct nfs_context *nfs, int enabled);
 EXTERN void nfs_set_dircache(struct nfs_context *nfs, int enabled);
 EXTERN void nfs_set_autoreconnect(struct nfs_context *nfs, int num_retries);
+EXTERN void nfs_set_retrans(struct nfs_context *nfs, int retrans);
 EXTERN void nfs_set_nfsport(struct nfs_context *nfs, int port);
 EXTERN void nfs_set_mountport(struct nfs_context *nfs, int port);
 EXTERN void nfs_set_readdir_max_buffer_size(struct nfs_context *nfs, uint32_t dircount, uint32_t maxcount);
@@ -1996,6 +1997,9 @@ EXTERN int nfs_get_poll_timeout(struct nfs_context *nfs);
  * int milliseconds : timeout to be applied in milliseconds (-1 no timeout)
  *                    timeouts must currently be set in whole seconds,
  *                    i.e. units of 1000
+ *
+ * Note: Prefer the mount option timeo=<int> to set the timeout over directly
+ *       calling nfs_set_timeout().
  */
 EXTERN void nfs_set_timeout(struct nfs_context *nfs, int milliseconds);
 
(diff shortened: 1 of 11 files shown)


sahlberg:
Causal first review looks mostly good.   I will review properly tonight or tomorrow.

One small thing that failed to spot earlier is that we shouldn't assume that stderr is available.
Can you change the LOG* functions to instead log to a file descriptor that the application sppecifies through
rpc|nfs_set_log_fd()  or something.   Default to -1 to indicate "no logging".

linuxsmiths:
> Causal first review looks mostly good. I will review properly tonight or tomorrow.
> 
> One small thing that failed to spot earlier is that we shouldn't assume that stderr is available. Can you change the LOG* functions to instead log to a file descriptor that the application sppecifies through rpc|nfs_set_log_fd() or something. Default to -1 to indicate "no logging".

Since RPC_LOG() was logging to stderr I made LOG() (and earlier TLS_LOG) also log to that.
In general I think logging needs more thought. How about we tackle that in a separate change.
Since we are a library, it's best to use the logger provided by the user, f.e., user may want to add timestamp and/or some tag to the logs.
Something like how gnutls does it - it asks for the logging functions and then calls them.
f.e.,
gnutls_global_set_log_function(), 
gnutls_global_set_audit_log_function()

sahlberg:
On Mon, 29 Apr 2024 at 16:48, linuxsmiths ***@***.***> wrote:
>
> Causal first review looks mostly good. I will review properly tonight or tomorrow.
>
> One small thing that failed to spot earlier is that we shouldn't assume that stderr is available. Can you change the LOG* functions to instead log to a file descriptor that the application sppecifies through rpc|nfs_set_log_fd() or something. Default to -1 to indicate "no logging".
>
> Since RPC_LOG() was logging to stderr I made LOG() (and earlier TLS_LOG) also log to that.
> In general I think logging needs more thought. How about we tackle that in a separate change.
> Since we are a library, it's best to use the logger provided by the user, f.e., user may want to add timestamp and/or some tag to the logs.
> Something like how gnutls does it - it asks for the logging functions and then calls them.
> f.e.,
> gnutls_global_set_log_function(),
> gnutls_global_set_audit_log_function()

Agree. Lets handle logging as a separate issue.

include/nfsc/libnfs.h
```diff
@@ -1996,6 +1997,9 @@ EXTERN int nfs_get_poll_timeout(struct nfs_context *nfs);
  * int milliseconds : timeout to be applied in milliseconds (-1 no timeout)
  *                    timeouts must currently be set in whole seconds,
  *                    i.e. units of 1000
+ *
+ * Note: Prefer the mount option timeo=<int> to set the timeout over directly
+ *       calling nfs_set_timeout().
```
sahlberg, line 2002: This is policy not function so doesn't belong in the header file.

sahlberg:
Merged, thanks

PR #461, opened 2024-04-30
Misc changes

linuxsmiths:
- Update nfs-pthreads examples to specify number of threads and not chunk size per thread. This is more practical and works well for very lareg file sizes too.
- Added multithreading and pthread detection to cmake.
- Misc compiler warning fixes in examples code which was left from last time.

Testing done:
- Validated nfs-pthreads-writefile and nfs-pthreads-async-writefile by transferring and checksumming 25GB file.


sahlberg:
Merged, thanks

PR #462, opened 2024-05-01
Avoid rpc_mutex deadlock

linuxsmiths:
rpc_set_error() holds the rpc_mutex so it cannot be called by any function which already holds the mutex. In this case rpc_write_to_socket() was calling rpc_set_error() resulting in a deadlock. This was detected when running the stress test suite where I continously keep killing the connectiong with "ss -K" while data xfer is on.

To solve this I defined a nolock version for both the rpc and nfs set error functions. Audited the code and fixed one more occurrence. The audit may not be complete so created "error checking" mutex to catch such locking violations.

Didn't use recursive mutex as those encourage bad programming practice and such bugs may silently get hidden.

Testing done:
With this fix ran 100+ iterations of 100GB file xfer while killing all connections every 5 seconds.

### DIFF
--- a/lib/init.c
+++ b/lib/init.c
@@ -296,29 +296,39 @@ void rpc_set_auxiliary_gids(struct rpc_context *rpc, uint32_t len, uint32_t* gid
 void rpc_set_error(struct rpc_context *rpc, const char *error_string, ...)
 {
         va_list ap;
-	char *old_error_string;
 
 #ifdef HAVE_MULTITHREADING
         if (rpc->multithreading_enabled) {
                 nfs_mt_mutex_lock(&rpc->rpc_mutex);
         }
 #endif /* HAVE_MULTITHREADING */
-        old_error_string = rpc->error_string;
+
         va_start(ap, error_string);
+	free(rpc->error_string);
 	rpc->error_string = malloc(1024);
 	vsnprintf(rpc->error_string, 1024, error_string, ap);
         va_end(ap);
 
 	RPC_LOG(rpc, 1, "error: %s", rpc->error_string);
 
-	if (old_error_string != NULL) {
-		free(old_error_string);
-	}
 #ifdef HAVE_MULTITHREADING
         if (rpc->multithreading_enabled) {
                 nfs_mt_mutex_unlock(&rpc->rpc_mutex);
         }
 #endif /* HAVE_MULTITHREADING */
+}
+
+void rpc_set_error_nolock(struct rpc_context *rpc, const char *error_string, ...)
+{
+	va_list ap;
+
+	va_start(ap, error_string);
+	free(rpc->error_string);
+	rpc->error_string = malloc(1024);
+	vsnprintf(rpc->error_string, 1024, error_string, ap);
+	va_end(ap);
+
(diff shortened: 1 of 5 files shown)


linuxsmiths:
@sahlberg, I've done basic audit, so if you get a chance pls audit to see if we are calling rpc_set_error()/nfs_set_error() in contexts already holding the rpc_mutex.

fyi, I'll be away travelling for next couple of days so my responses may be delayed.

lib/init.c
```diff
@@ -296,31 +296,41 @@ void rpc_set_auxiliary_gids(struct rpc_context *rpc, uint32_t len, uint32_t* gid
 void rpc_set_error(struct rpc_context *rpc, const char *error_string, ...)
 {
         va_list ap;
-	char *old_error_string;
```
sahlberg, line 299: I don't think we can do this change and drop old_error_string.
The reason is it is useful to be able to append the previous error strong to the new one being generated.

For example instances like this:
                smb2_set_error(smb2, "Failed to create set command. %s",
                               smb2_get_error(smb2));

So undo this change.

lib/init.c
```diff
@@ -296,31 +296,41 @@ void rpc_set_auxiliary_gids(struct rpc_context *rpc, uint32_t len, uint32_t* gid
...
-		free(old_error_string);
-	}
 #ifdef HAVE_MULTITHREADING
         if (rpc->multithreading_enabled) {
                 nfs_mt_mutex_unlock(&rpc->rpc_mutex);
         }
 #endif /* HAVE_MULTITHREADING */
 }
 
+void rpc_set_error_nolock(struct rpc_context *rpc, const char *error_string, ...)
```
sahlberg, line 321: See the previous comment on why we need the old_error_string variable here and why we need to defer freeing it until after we have created the new string in vsnprintf.

lib/init.c
```diff
@@ -296,31 +296,41 @@ void rpc_set_auxiliary_gids(struct rpc_context *rpc, uint32_t len, uint32_t* gid
...
 
 #ifdef HAVE_MULTITHREADING
         if (rpc->multithreading_enabled) {
                 nfs_mt_mutex_lock(&rpc->rpc_mutex);
         }
 #endif /* HAVE_MULTITHREADING */
-        old_error_string = rpc->error_string;
+
         va_start(ap, error_string);
+	free(rpc->error_string);
```
sahlberg, line 307: Ideally this whole function should just be the locking and then a call to the _locked() function
so that we only have one copy of the code that generates the new error string.

lib/init.c
sahlberg, line 321: Change the name to rpc_set_error_locked() to indicate this is called when the lock is already held.

sahlberg (changes requested):

lib/init.c
linuxsmiths: good point! will update

lib/init.c
linuxsmiths: yes, I thought too but due to the varargs we have to either take a va_list as argument in which case the locked() version cannot be directly called or we make that a macro. For now, I'll leave it as is.

PR #463, opened 2024-05-01
Misc improvements in the nfs-pthreads-*writefile examples

linuxsmiths:
- Use 1MB write size
- Limit max outstanding in case of async writes, else it can grow unbounded for very large files

### DIFF
--- a/examples/nfs-pthreads-async-writefile.c
+++ b/examples/nfs-pthreads-async-writefile.c
@@ -78,6 +78,12 @@ WSADATA wsaData;
  */
 #define NUM_THREADS  8
 
+/*
+ * Max how many writes can be outstanding at any tine.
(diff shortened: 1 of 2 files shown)


examples/nfs-pthreads-async-writefile.c
```diff
@@ -78,6 +78,12 @@ WSADATA wsaData;
  */
 #define NUM_THREADS  8
 
+/*
+ * Max how many writes can be outstanding at any tine.
+ * This will be divided by NUM_THREADS to decide per-thread outstanding.
+ */
+#define MAX_CONCURRENCY 1024
```
sahlberg, line 85: That is a low number but ok.   Many servers handle up to several hundreds of thousand concurrent requests.

sahlberg (approved):

PR #476, opened 2024-07-20
Add mount options - rsize, wsize, and set readmax/writemax and readdr maxcount/dircount correctly per FSINFO negotiation and user selected values.

linuxsmiths:


sahlberg:
Merged. Thanks!
The PlayStation2 build failures I think are unrelated. I think it is just due to updated sdk and compilers for that environment.
I will fix that tomorrow or so.

PR #477, opened 2024-07-21
Added rpc_stats for maintaining stats for each rpc_context

linuxsmiths:
This is required since RPC layer performs retransmits, timeout handling, reconnect, etc, so user won't know those. get_rpc_stats() is the new API added for fetching stats for an rpc contect.

### DIFF
--- a/include/libnfs-private.h
+++ b/include/libnfs-private.h
@@ -205,6 +205,8 @@ struct tls_context {
 };
 #endif /* HAVE_TLS */
 
+#define INC_STATS(rpc, stat) ++((rpc)->stats.stat)
+
 struct gss_ctx_id_struct;
 struct rpc_context {
 	uint32_t magic;
@@ -360,6 +362,9 @@ struct rpc_context {
         /* Is a server context ? */
         int is_server_context;
         struct rpc_endpoint *endpoints;
+
+        /* Per-transport RPC stats */
+        struct rpc_stats stats;
 };
 
 struct rpc_pdu {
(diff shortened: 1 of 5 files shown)


lib/socket.c
```diff
@@ -810,6 +812,9 @@ rpc_timeout_scan(struct rpc_context *rpc)
 			continue;
 		}
 
+		// Timed out w/o being sent.
```
sahlberg, line 815: please use c-style comments,   i.e. /* */

include/libnfs-private.h
```diff
@@ -205,6 +205,8 @@ struct tls_context {
 };
 #endif /* HAVE_TLS */
 
+#define INC_STATS(rpc, stat) ++((rpc)->stats.stat)
```
sahlberg, line 208: Should this use atomic_fetch_add_explicit() ?
Right now I think all callpaths are guaranteed to be from the same thread  but future updates and additional stats might be added from different threads and this could become racy, unless the multithreading mutex is held.

linuxsmiths: Let's not add atomic till we need it. Since these are RPC stats the updates to them would be done by a single thread and in other places it'll be already protected by the rpc mutex.

lib/socket.c
linuxsmiths: Updated

include/libnfs-private.h
linuxsmiths: Also added a note to call this out in case we add more counters in future.

linuxsmiths: Also, the counters are updated correctly only for clients. I see some code to allow libnfs based servers, but I'm not sure how complete is that.

sahlberg:
Merged. Thanks!

include/libnfs-private.h
sahlberg: We can fix that later.
There are a couple of severs built on libnfs I am aware of but they are all for internal dedicated use in some processing workflows and not general purpose of public.

PR #498, opened 2024-11-20
initialize iov_ref for nfs4

ggoers:
**Problem**: 
In commit [ae86b44](https://github.com/sahlberg/libnfs/commit/ae86b44dbb93840451151e0513230f1d7c18a263), the iov_ref was introduced. However, the initialization was done only for nfs3 and not for nfs4. As a result the assertions added in the same commit caused failures when trying to read a file with nfs4.


sahlberg:
Merged. Thanks!

PR #508, opened 2024-12-16
Fix win32 build of installed package

### LINKED ISSUE #507: libnfs-raw.h includes files that are not installed
`libnfs-raw.h` is installed with the library (used in VLC). But it includes files that are not installed like `win32_compat.h`: https://github.com/sahlberg/libnfs/blob/2e7971e1b97454e309eca7a74305fa8ee5efc943/include/nfsc/libnfs-raw.h#L43

This makes it impossible to build for Windows with the installed package.

robUx4:
`win32_compat.h` is not installed so it should not be included from an installed header.

Also includes other fixes for mingw-w64 and MSVC.

Fixes #507

### DIFF
--- a/include/nfsc/libnfs-raw.h
+++ b/include/nfsc/libnfs-raw.h
@@ -39,10 +39,6 @@
 #include "ps3_compat.h"
 #endif
 
-#ifdef WIN32
-#include <win32/win32_compat.h>
-#endif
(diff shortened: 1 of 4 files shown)


sahlberg:
Merged. Thanks.
Very much appreciated.

PR #509, opened 2024-12-16
win32: don't use pthread on Windows

### LINKED ISSUE #506: 6.0.2 fails to build in mingw-w64
`mingw-w64` has winpthreads support. That means the autotools/CMake detect that pthread is available (although you need to link with `-lpthreads`). The libnfs code use Windows API for threading (good) but the code is not exclusive to `HAVE_PTHREAD`, so some threading code is defined twice. Some types are also defined twice: https://github.com/sahlberg/libnfs/blob/2e7971e1b97454e309eca7a74305fa8ee5efc943/include/libnfs-multithreading.h#L51

The `WIN32` and `HAVE_PTHREAD` code should be exclusive.

robUx4:
mingw-w64 has winpthread which is detected as available. But on Windows there is custom threading code to use. We should only use one of them. And the win32 API doesn't require an extra -lpthread to link.

Fixes #506

### DIFF
--- a/include/libnfs-multithreading.h
+++ b/include/libnfs-multithreading.h
@@ -29,7 +29,12 @@ extern "C" {
 
 #ifdef HAVE_MULTITHREADING
 
-#ifdef HAVE_PTHREAD
+#ifdef WIN32
+typedef HANDLE libnfs_thread_t;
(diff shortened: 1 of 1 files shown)


sahlberg:
Merged. Thanks!
Very much appreciated!

PR #510, opened 2024-12-16
cmake: export the necessary library in the pkg-config file

robUx4:
That includes gnutls on Linux if it's used.


sahlberg:
Merged. Thanks

PR #518, opened 2025-01-26
move link against gnutls into main library, as it is referenced by tls/libtls.la

fundawang:
As titled.

FYI, I could only use the web interface of github, as github.com is banned here in China, so I could not clone nor push it via command. So there are two separarated commits.

### DIFF
--- a/lib/Makefile.am
+++ b/lib/Makefile.am
@@ -48,5 +48,5 @@ libnfs_la_LIBADD = \
 
 if HAVE_TLS
 libnfs_la_CPPFLAGS += -I$(abs_top_srcdir)/tls
-libnfs_la_LIBADD += ../tls/libtls.la
+libnfs_la_LIBADD += ../tls/libtls.la -lgnutls
 endif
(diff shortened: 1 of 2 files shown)


sahlberg:
> As titled.
> 
> FYI, I could only use the web interface of github, as github.com is banned here in China, so I could not clone nor push it via command. So there are two separarated commits.

You can also "git format-patch" in a local repo and just email it to me.

fundawang:
> > As titled.
> > FYI, I could only use the web interface of github, as github.com is banned here in China, so I could not clone nor push it via command. So there are two separarated commits.
> 
> You can also "git format-patch" in a local repo and just email it to me.

No, it is not feasible, as I could not `git clone` :(

sahlberg:
> No, it is not feasible, as I could not `git clone` :(

That sucks. Do you have any suggestions on alternative free-to-use git repositories where I can host a mirror of it?

fundawang:
> > No, it is not feasible, as I could not `git clone` :(
> 
> That sucks. Do you have any suggestions on alternative free-to-use git hosting sites where I can host a mirror of it?

github just fine. There is nothing we can do about this.

PR #514, opened 2025-01-14
socket: use void cast to allow compile for arm32

heitbaum:
When compiling for arm32 the error: cast increases required alignment of target type occurs.

/lib/socket.c: In function 'rpc_read_from_socket': lib/socket.c:586:49: error: cast increases required alignment of target type [-Werror=cast-align]
  586 |                         rpc->rm_xid[1] = ntohl(*(uint32_t *)&buf[0]);
      |                                                 ^
lib/socket.c:587:61: error: cast increases required alignment of target type [-Werror=cast-align]
  587 |                         rpc->pdu = rpc_find_pdu(rpc, ntohl(*(uint32_t *)&buf[0]));
      |                                                             ^
lib/socket.c:799:35: error: cast increases required alignment of target type [-Werror=cast-align]
  799 |                                 *((uint32_t *)rpc->inbuf) = htonl(rpc->rm_xid[1]);
      |                                   ^
lib/socket.c:861:71: error: cast increases required alignment of target type [-Werror=cast-align]
  861 |                                                 const READ3res *res = (READ3res *) rpc->pdu->zdr_decode_buf;
      |                                                                       ^

cast via (void *) to allow the compile to suceed, as all of the source buffers should be 32bit aligned.

- fixes #513 
- Fixes https://github.com/LibreELEC/LibreELEC.tv/issues/9678

Build tested on LibreELEC with both Raspberry RPi2 and Allwinner H3 targets.

PR #522, opened 2025-02-19
Fix cast-align for riscv64

cybaol:
Part3 of #514 and #519


sahlberg:
Merged, thanks!

kxxt:
The `void* `cast added in this PR shouldn't be necessary, since #519 already turned the type of `zdr_decode_buf` from effectively `char *` into `void *`.

PR #527, opened 2025-03-05
Adapt RFC documents when using AUTH_UNIX

JiaLong-Kang:
Hi Ronnie,

When we use AUTH_UNIX, the number of auxiliary groups should be limited, refer to [RFC 5531 section 14](https://datatracker.ietf.org/doc/html/rfc5531#page-25).


JiaLong-Kang:
If the number of auxiliary groups is greater than 16, we only send the first 16 groups, and we need to tell the user this through some mechanism. 

I think returning NULL is the most direct way to notify.

sahlberg:
Merged. thanks!

PR #529, opened 2025-03-06
Fix compilation failure when multi-threading option is enabled

JiaLong-Kang:
Hi Ronnie,

Compilation fails when multithreading is enabled.

Environment description:
platform: centos7
gcc version: 9.3.1
 

截屏2025-03-06 18 40 25

`config.h` file automatically generated by cmake
image


sahlberg:
Merged, thanks!

PR #531, opened 2025-03-07
Add nfs4_getattr function

zhiburt:
Hi @sahlberg 

It was noticed that `nfs_get_writemax` does not work properly with NFSv4.
Which is in my opinion correct cause protocol is pretty different and it would require unnecessary rpc calls, which not everybody needs.

This patch adds a function to obtain a proper attributes from server.

Function declaration intentionally made more generic then ought to, so more attributes could be added later on with no change to interface.

```c
struct nfs_context *nfs;
fattr4_attr nfsattr;

nfs4_getattr(nfs, FATTR4_ATTR_TYPE_MAXREAD | FATTR4_ATTR_TYPE_MAXWRITE, &nfsattr);
```

Let me know if I got something wrong.
Take care.

### DIFF
--- a/include/nfsc/libnfs.h
+++ b/include/nfsc/libnfs.h
@@ -309,11 +309,17 @@ EXTERN const struct sockaddr_storage *nfs_get_server_address(struct nfs_context 
 
 /*
  * Get the maximum supported READ size by the server
+ *
+ * Notice that in case of NFSv4 it may not get you a proper configuration of server.
+ * You can call nfs4_getattr to retrieve it.
  */
 EXTERN size_t nfs_get_readmax(struct nfs_context *nfs);
 
 /*
  * Get the maximum supported WRITE size by the server
+ *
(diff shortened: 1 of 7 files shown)


sahlberg:
I don't think we need a new function with the too generic name nfs4_getattr()

We have everything already except requesting 
const FATTR4_MAXREAD            = 30;
const FATTR4_MAXWRITE           = 31;

These two parameters I think would fit conceptualy in the statvfs structure.

I think a better approach is to :
1, add new fields maxread, maxwrite to the nfs_statvfs_64 structure
2, add FATTR4_MAXREAD/FATTR4_MAXWRITE to the requested stats in statvfs_attributes
3, add parsing of them to nfs_parse_statvfs64

sahlberg:
Even better way is to just read and update these limits during the normal mount process.

I have updated master to update these limits for nfsv4 so it should work just like ot does for v3 now.

thanks for the report.

zhiburt:
Good;
So I guess it can be closed then?

PR #534, opened 2025-03-14
Allow getexports to use a different mountport

afbjorklund:
This is so that args such as mountport can be used, previously it would ignore them and always use 111.

### DIFF
--- a/include/libnfs-private.h
+++ b/include/libnfs-private.h
@@ -350,6 +350,8 @@ struct rpc_context {
 	 */
 	char *server;
 
+	int mountport;
+
 	/* fragment reassembly */
 	struct rpc_fragment *fragments;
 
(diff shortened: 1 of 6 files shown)


sahlberg:
Let's not change these signatures.
Instead, you can add the mountport to the rpc_context structure and add getters and setters for it.

The nfs context is not appropriate for this as the exports, mount protocol, is not really part of the nfs protocol.

afbjorklund:
Added a copy of mountport to rpc_context, and added a mountport parameter to mount_getexports.

```console
$  ./utils/nfs-ls -D "nfs://localhost?nfsport=40479&mountport=40479"
nfs://localhost/mount
$ ./utils/nfs-ls -D nfs://localhost:40479
nfs://localhost/mount
```

sahlberg:
Merged,  thanks!

PR #538, opened 2025-03-23
Allocate memory for correct number of ops

GlenWalker:
- 1 for PUTFH or PUTROOTFH
- 1 LOOKUP for each path component
- 1 for GETATTR
- Additional as specified with num_extra by caller

* This calculation has allocated 2 ops per path component since 3ced39f5
* Perhaps there was originally a plan to perform 2 ops per path component, LOOKUP and GETATTR, for filehandle caching purposes, that was not completed


sahlberg:
Merged. Thanks!

PR #539, opened 2025-03-23
Fix cmakelists for Windows utilities

soffensive:
Fix the build and compilation process for Windows

We don't need to link against the nfs_mount library, but only the nfs library.
However, we need the mount header files for the compilation (therefore target_include_directories)


sahlberg:
Merged, thanks

PR #540, opened 2025-03-23
Explicit shared library output filename for consistency across platforms

soffensive:
Be explicit about the filename of the output shared library For Windows there will be no automatic "lib" prefix, so we ensure a consistent file name across platform


sahlberg:
Merged, thanks!

PR #544, opened 2025-04-10
Fix use-after-free error

GlenWalker:
Both check_nfs4_error and nfs4_find_op will free data if an error is found, so accessing data->cb afterwards leads to a segfault. This change makes nfs4_getacl_cb consistent with other functions in nfs_v4.c that simply return if check_nfs4_error or nfs4_find_op return value indicates an error.


sahlberg:
Merged. Thanks!

Issue #547, opened 2025-04-11
Potential memory leak in nfs_parse_url_dir


sg3-141-592:
I've been experimenting with fuzzing the different libnfs url parsers and I think it's found a potential memory leak.

For example calling `nfs_parse_url` with url "nfs://@#/##"

`nfs_parse_url` allocs this string.

https://github.com/sahlberg/libnfs/blob/fdba371a1c95167dd4f2924049683255721eda18/lib/libnfs.c#L392

Later on if it's a url with `@` in, it re-allocs `urls->server` without free-ing the earlier allocation.

https://github.com/sahlberg/libnfs/blob/fdba371a1c95167dd4f2924049683255721eda18/lib/libnfs.c#L524-L531

This is how you'd re-create 

```
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include "nfsc/libnfs.h"

int main(int argc, char *argv[])
{
    struct nfs_url *nfs_url;
    struct nfs_context *nfs;

    nfs = nfs_init_context();
    if (nfs == NULL) {
        fprintf(stderr, "failed to init context\n");
        return 0;
    }

    nfs_url = nfs_parse_url_full(nfs, "nfs://@#/##");

    if (nfs_url != NULL) {
        nfs_destroy_url(nfs_url);
    }
    
    nfs_destroy_context(nfs);

    return 0;
}
```

```
$ valgrind --leak-check=full ./fuzz/recreate_leak 
......
==83492== 6 bytes in 1 blocks are definitely lost in loss record 1 of 54
==83492==    at 0x484880F: malloc (vg_replace_malloc.c:446)
==83492==    by 0x4B8558E: strdup (strdup.c:42)
==83492==    by 0x487899C: nfs_parse_url (lib/libnfs.c:378)
==83492==    by 0x48788F0: nfs_parse_url_full (lib/libnfs.c:548)
==83492==    by 0x1091DD: main (fuzz/recreate_leak.c:17)
==83492== 
==83492== LEAK SUMMARY:
==83492==    definitely lost: 6 bytes in 1 blocks
```

And from libfuzzer and Clang [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) that I used to find the issue 

```
=================================================================
==86369==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 11 byte(s) in 1 object(s) allocated from:
[...]

PR #550, opened 2025-04-12
Fix memory leak around username handling

sg3-141-592:
Fix for memory leak in `nfs_parse_url` where a value of `urls->server` is not freed when handling a URL with an `@` in it #547 .

I've added a test case for parsing urls via `nfs_parse_url_full` that can capture this leak. I couldn't get a working test case for urls with port numbers. It looks like it tries to cast everything after the `:` in a url as int, but that doesn't work for `nfs://127.0.0.1:8000/dir/file` style paths.

https://github.com/sahlberg/libnfs/blob/fdba371a1c95167dd4f2924049683255721eda18/lib/libnfs.c#L434-L438

### DIFF
--- a/lib/libnfs.c
+++ b/lib/libnfs.c
@@ -375,7 +375,7 @@ static struct nfs_url *
 nfs_parse_url(struct nfs_context *nfs, const char *url, int dir, int incomplete)
 {
 	struct nfs_url *urls;
-	char *strp, *flagsp, *strp2, ch;
+	char *strp, *flagsp, *strp2, ch, *original_server;
         int tmp;
 
 	if (strncmp(url, "nfs://", 6)) {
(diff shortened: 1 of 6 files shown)


sahlberg:
Nice.
Remove the "DRAFT" tag and I will merge it.  Thanks!

lib/libnfs.c
```diff
@@ -528,14 +528,17 @@ nfs_parse_url(struct nfs_context *nfs, const char *url, int dir, int incomplete)
                         nfs_destroy_url(urls);
                         return NULL;
                 }
+                char *original_server = urls->server;
```
sahlberg, line 531: Move the declaration to the head of the function. I try not ti intermingle code and variable declarations.

sahlberg (changes requested):

lib/libnfs.c
sg3-141-592: `*original_server` moved up to head of function

sahlberg:
Merged. Thanks!

PR #552, opened 2025-04-14
Adding parsing of port numbers

sg3-141-592:
As part of doing #550 I noticed that port number parsing isn't working correctly. Currently the `nfs_parse_url` assumes that the rest of string is a port number after seeing a `:`. This doesn't work for cases where there is more to parse e.g. `nfs://localhost:8000/dir/file`.

https://github.com/sahlberg/libnfs/blob/fdba371a1c95167dd4f2924049683255721eda18/lib/libnfs.c#L434-L438

- `strcpy(strp + 1, strp + 3);` overlapping `strcpy` operations are technically undefined. I've switched it to `memmove`.

### DIFF
--- a/lib/libnfs.c
+++ b/lib/libnfs.c
@@ -375,8 +375,9 @@ static struct nfs_url *
 nfs_parse_url(struct nfs_context *nfs, const char *url, int dir, int incomplete)
 {
 	struct nfs_url *urls;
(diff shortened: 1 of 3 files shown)


sahlberg:
That is a good find. Thanks for also adding a test for it.

PR #557, opened 2025-05-22
Don't overwrite POLLERR error with less specific POLLHUP error

GlenWalker:
If connection fails both POLLERR and POLLHUP are triggered, but the error message from POLLERR is more useful as it includes the socket error number, while the error message from POLLHUP is very generic


sahlberg:
Merged. Thanks!

PR #562, opened 2025-06-21
Use correct buffer size of ZDR to avoid memory corruption

GlenWalker:
ZDR is being initialised with a buffer size that is 4 bytes larger than the buffer it is actually given (as the first 4 bytes have been used for IOV). This allows libnfs to write 4 bytes past the end of the buffer and cause memory corruption instead of returning an error.


sahlberg:
Merged, thanks!

PR #584, opened 2026-03-23
Fix linking of libnfs to libpthread

brad0:
libnfs utilizes a bunch of POSIX threads functions but does
not link against libpthread as it should nor the pkg-config file
properly listing libpthread as a static link time dependency.


sahlberg:
Merged. Thanks

PR #592, opened 2026-07-27
socket: stop reading when a callback replaces the connection

nebulabox:
A reply callback may disconnect the current socket and start an asynchronous connection to another endpoint on the same RPC context. This happens after a portmapper reply when switching to the target RPC service.

The read loop could continue and call recv() on the replacement socket before its connection completed, resulting in WSAENOTCONN on Windows.

Return to the event loop when a client context is no longer connected, allowing poll() to complete the replacement connection before further I/O.

### DIFF
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -654,6 +654,18 @@ rpc_read_from_socket(struct rpc_context *rpc)
 	}
 
         while (1){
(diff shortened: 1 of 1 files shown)


sahlberg:
Merged, thanks!

PR #594, opened 2026-08-04
Fixing RPC disconnect after reply callback

jaypitts23:
A reply callback (e.g. the portmap GETPORT handler) can disconnect the socket we are reading on and open a new connection while we are still inside this read loop.


sahlberg:
Merged. Thanks!

PR #609, opened 2026-09-09
nfs4: retry replay-safe NFS4ERR_DELAY replies

kogeler:
## Summary

- Retry `NFS4ERR_DELAY` asynchronously instead of immediately failing requests such as `OPEN` and `RENAME`. Limit this to ordinary NFSv4.2 `AUTH_SYS` compounds where `SEQUENCE` succeeded and no earlier state-changing operation succeeded.
- Release the session slot while waiting, then retry with a new RPC XID and sequence number while preserving the original arguments and credentials. Allow at most eight retries, with backoff from 100 ms to 1 s. Pending retries take part in queue accounting, cancellation and context cleanup.
- Leave partial mutations, errors from `SEQUENCE`, NFSv4.0, GSS and zero-copy I/O unchanged. This is not a general session recovery implementation.

## Test plan

- [x] Reproduced the original failure with a deterministic RPC peer: `OPEN` returned status 10008 instead of succeeding after a retry.
- [x] Ran all 11 new regression cases, including repeated DELAY, concurrent requests, real I/O and permission errors, partial success, retry exhaustion and cancellation.
- [x] Ran the RPC fixture and modified library units with ASan.
- [x] Tested through fuse-nfs against an NFSv4.2 server: eight fio workers for at least 600 seconds, CRC32C readback and buffered fsync, with no errors.
- [ ] The full existing upstream test suite was not run.

The regression test is included as `tests/prog_nfs4_delay.c` and was compiled and run separately. It is distributed through `EXTRA_DIST`, but is not yet wired into `make check` or CTest.


sahlberg:
merged with some minor fixes.   Thanks!

Issue #612, opened 2026-09-13
libFuzzer harness for the RPC reply decoders


youhaveme9:
Hi Ronnie,

I've been fuzzing the reply decoding path in libnfs (the client side, everything a server sends back) and would like to contribute the harness, with the eventual goal of getting libnfs onto OSS Fuzz so it runs there continuously.

The harness drives the generated ZDR decoders directly, the same dispatch rpc_process_pdu uses. The first input byte picks the decoder and the rest of the input is the reply payload. All of them are covered, 157 decoders across NFS v2/v3, mount, NLM, portmap, rquota and the NFSv4.x compound. A short ASan campaign over the seed corpus reaches deep into the table with no findings, which matches how well the code has held up since the security scan.

OSS Fuzz wants new integrations to keep the fuzz target in the upstream repo with the maintainer aware, hence asking first: would you take this in tree? I'd put it in fuzzing/ or tests/fuzzing/, standalone, just the harness, a seed corpus and a README with the clang one liner to build it. No changes to the rest of the build.

If that works I'll send a PR. Also fine with a different location or approach if you prefer.

sahlberg:
Sure, send a PR and I will have a look.

sahlberg:
Merged.

youhaveme9:
Hi Ronnie, one small follow up on the fuzzing work. The integration PR on Google's side (google/oss-fuzz#16134) is in final review, and the reviewer would like a libnfs maintainer to confirm on that PR that the contact email is right. The project currently lists ronniesahlberg@gmail.com as the address where reports of any issues the continuous fuzzing finds would be sent, with my address in copy. A short comment from you on the Google PR confirming that address works would be all it needs.

PR #613, opened 2026-09-14
Add a libFuzzer target for RPC reply decoding

youhaveme9:
As discussed in #612.

The harness feeds a buffer to one of the generated ZDR reply decoders, picked by the first input byte, the same dispatch rpc_process_pdu uses. All 157 decoders are covered: NFS v2/v3, mount v1/v3, NLM, portmap v2/v3, rquota and the NFSv4.x compound.

Everything is standalone under `fuzzing/`, nothing else in the tree changes. Once this is in I'd like to hook libnfs up to OSS Fuzz so it keeps running under sanitizers there.


sahlberg:
Merged. Thanks!

PR #614, opened 2026-09-15
Map NFSv4 stale filehandles to ESTALE

kogeler:
## Summary

- Return `-ESTALE` for `NFS4ERR_STALE`, matching the existing NFSv3 mapping. `NFS4ERR_IO` still returns `-EIO`.
- I ran into this in fuse-nfs when another client replaced a file and a cached lookup still held its old server filehandle. Reporting this as an I/O error hides the distinction the adapter/VFS needs to revalidate a stale name. This does not recover a file the server has already removed.
- Add a small regression covering STALE and the unchanged success, IO, NOENT and ACCESS mappings.

## Test plan

- [x] CMake Debug build and the four non-mount unit tests.
- [x] The same tests with ASan/UBSan and leak detection.
- [x] Autotools build and `make -C tests check`.
- [x] Negative control: the regression fails on the original STALE-to-EIO mapping.


sahlberg:
Merged. Thanks!

